PR #96, opened 2026-06-22
fix(dev-container): install Docker Buildx so deploy-all works on Apple Silicon

mvinci12:
## Problem

Running `gco stacks deploy-all` from the dev container on an arm64 host (Apple Silicon) fails at two separate steps:

1. **Volcano ECR image mirror** aborts before CloudFormation:
   ```
   Image mirror failed for region <r>: No multi-arch image-copy method available.
   Need one of: 'docker buildx' (Docker Buildx), 'docker pull --all-platforms'
   (Finch/nerdctl), or skopeo on PATH.
   ```
2. **`linux/amd64` asset build** (the Lambda/service images CDK bundles) fails:
   ```
   docker build ... --platform linux/amd64 ... exited with error code 1:
   ... image ... does not provide the specified platform (linux/amd64)
   ```

The dev container is the **documented, supported path** for deploying stacks (README/QUICKSTART), and the docs already state the deploy needs *Docker Buildx, Finch, or skopeo* (`docs/IMAGE_MIRROR.md`, `docs/CUSTOMIZATION.md`) — but `Dockerfile.dev` ships **none** of them. The ECR image mirror feature added in v3.0.0 (#83) introduced this requirement without adding the tooling to the image.

## Root cause

- The Docker CLI install in `Dockerfile.dev` extracts only the `docker` client binary from the static tarball, which does **not** include `cli-plugins/` (no Buildx).
- On arm64, the legacy builder cannot cross-build `linux/amd64`, and there is no multi-arch copy tool for the mirror.

## Fix

Install the **Docker Buildx** CLI plugin in `Dockerfile.dev`, pinned (`v0.35.0`) and selected by `$TARGETARCH`, into a system `cli-plugins` path.

Buildx alone satisfies **both** steps:
- It is the preferred mirror strategy in `cli/_image_mirror.py::resolve_copy_strategy()` (`docker buildx imagetools create`).
- It provides BuildKit cross-builds for the `linux/amd64` assets.

skopeo/Finch were intentionally not added — Buildx is sufficient and is the project's first-listed preferred tool.

Docs updated: README, QUICKSTART, and a new TROUBLESHOOTING entry mapping both error strings to the fix.

## Verification (local, no AWS spend)

- Native arm64 rebuild of `Dockerfile.dev` succeeds; `docker buildx version` → `v0.35.0`, `gco --version` → `3.2.2`.
- `docker build --platform linux/amd64` inside the image uses BuildKit (no legacy/DEPRECATED warning) and produces an `amd64` image.
- `markdownlint-cli2` passes (0 errors) on the changed docs.
- Originally surfaced during a real multi-region `deploy-all` (us-west-2 + ap-southeast-1); both failures reproduced and resolved by adding Buildx.

## Files changed

- `Dockerfile.dev` — install Buildx plugin + header note
- `README.md`, `QUICKSTART.md` — mention Buildx in the dev-container tooling list
- `docs/TROUBLESHOOTING.md` — new entry for the two error strings

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -5,7 +5,10 @@
 # Docker Desktop on Intel Macs / Linux) and linux/arm64 (Apple Silicon
 # Macs running Docker Desktop, Finch, Colima, or any arm64 Linux host).
 # kubectl, the AWS CLI v2, and the Docker static client all publish both
-# arch builds; we pick the right one via BuildKit's $TARGETARCH.
+# arch builds; we pick the right one via BuildKit's $TARGETARCH. The
+# Docker Buildx plugin is installed alongside the client so the multi-arch
+# image mirror and the linux/amd64 asset builds that `gco stacks deploy-all`
+# performs work out of the box on Apple Silicon (arm64) as well as x86_64.
 #
 # Build:
 #   docker build -f Dockerfile.dev -t gco-dev .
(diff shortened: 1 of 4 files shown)


Jmevorach:
Thank you @mvinci12 ; really appreciate this contribution! This is great. Will fix this shortly.

Jmevorach:
This PR’s work is now in v3.3.0 🎉!

PR #126, opened 2026-07-06
fix: make regional auth-secret IAM ARN deterministic for single-regio…

### LINKED ISSUE #125: Single-region deploy fails cdk-nag IAM5 on ServiceAccountRole auth secret
## Summary
In a single-region topology (global/api_gateway stacks co-located with the regional stack), `cdk synth`/`gco stacks deploy` fails cdk-nag validation with an unsuppressed `AwsSolutions-IAM5` wildcard finding on the regional `ServiceAccountRole` auth-secret policy.

## Environment
- GCO version: 3.8.5
- Python version: 3.14.6
- AWS CLI version: aws-cli/2.33.3 Python/3.13.11 Darwin/25.5.0 exe/arm64
- CDK CLI version: 2.1118.0 (build 2a848bc)
- Node.js / npm version: v24.9.0 / 11.6.0
- OS: macOS 26.5.1 (also reproducible in the project dev container)
- AWS region(s) affected: any single-region setup where `deployment_regions.global` (and `api_gateway`) equals a `regional` entry (e.g., all `us-east-1`)
- Kubernetes version (if applicable): N/A — fails at synth, before any cluster interaction

## Steps to reproduce
1. In `cdk.json`, set `deployment_regions` so the global stacks share the regional region:
   ```json
   "deployment_regions": {
     "global": "us-east-1",
     "api_gateway": "us-east-1",
     "monitoring": "us-east-1",
     "regional": ["us-east-1"]
   }
   ```
2. Run `gco stacks deploy gco-us-east-1 -y` (or `cdk synth gco-us-east-1`).
3. Synthesis reaches the cdk-nag "Performing Policy Validations" stage.

## Expected behavior
Synthesis succeeds. The `ServiceAccountRole` wildcard permission on the API Gateway auth secret is covered by an existing `AwsSolutions-IAM5` suppression, exactly as it is in the default (cross-region) topology.

## Actual behavior
Synthesis fails with an unsuppressed IAM5 finding on `gco-us-east-1/ServiceAccountRole/DefaultPolicy`. The finding detail references the cross-stack export token for the auth secret, which is not present in the suppression list.

```text
ERROR The IAM entity contains wildcard permissions and does not have a cdk-nag
rule suppression with evidence for those permission. ... The findings are in the
format 'Action::<action>' for policy actions and 'Resource::<resource>' for
resources. Example: appliesTo: ['Action::s3:*']. (AwsSolutions)
gco-us-east-1/ServiceAccountRole/DefaultPolicy/Resource aws-cdk-lib.aws_iam.CfnPolicy
Acknowledge with 'AwsSolutions::AwsSolutions-IAM5[Resource::gco-api-gateway:ExportsOutputRefGCOAuthSecret<hash>*]'
Synthesis finished with errors
✗ Deployment failed
```

## Additional context
Root cause: the regional `ServiceAccountRole` policy in `gco/stacks/regional_stack.py` grants the auth secret with a wildcard resource:

```python
resources=[f"{self.auth_secret_arn}*"]  # Wildcard to match with or without suffix
```

`self.auth_secret_arn` is `api_gateway_stack.secret.secret_arn`, a cross-stack reference token. How it renders into the IAM `Resource` depends on topology:
- Cross-region (default `global=us-east-2`): renders as the literal ARN `arn:aws:secretsmanager:<region>:<account>:secret:gco/api-gateway-auth-token*`, which **is** in the `add_iam_suppressions` list in `gco/stacks/nag_suppressions.py`.
- Same-region: CDK resolves it as a native cross-stack export/import, so the detail becomes `gco-api-gateway:ExportsOutputRefGCOAuthSecret<hash>*`, which is **not** in the list.

Because cdk-nag v3 matches finding details verbatim (no regex), the same-region form is unsuppressed and synth errors. The default `cdk.json` ships `global: us-east-2`, which masks this in the out-of-the-box config; it only surfaces when an operator collapses to a single region.

Suggested fix (not yet implemented): make the IAM resource ARN deterministic instead of deriving it from the volatile cross-stack token — e.g. build it with `Stack.format_arn(...)` from `project_name` + the global region so it renders identically in both topologies (`arn:aws:secretsmanager:<region>:<AWS::AccountId>:secret:<project>/api-gateway-auth-token-*`). The suppression can then be derived from `project_name`/region rather than a synthesis-time logical-id hash. This preserves the intentional suffix wildcard while removing the topology dependency. A regression test that synths both same-region and cross-region configs would prevent recurrence.

## Checklist
- [ ] I searched existing issues and this is not a duplicate
- [x] I removed any secrets, account IDs, or PII from logs pasted above
- [x] I can reproduce this consistently, or I've described the conditions under which it occurs

Jmevorach:
…n deploys

The regional ServiceAccountRole granted read on the API Gateway auth secret using `f"{self.auth_secret_arn}*"`, where `auth_secret_arn` is the cross-stack `api_gateway_stack.secret.secret_arn` token. That token renders differently by topology: a literal ARN cross-region (covered by the AwsSolutions-IAM5 suppression) but a native cross-stack export (`gco-api-gateway:ExportsOutputRefGCOAuthSecret<hash>`) when the API Gateway stack is co-located with the regional stack. The export form was never in the suppression list, so `cdk synth` / `gco stacks deploy` failed cdk-nag IAM5 on any single-region topology.

Build the IAM resource as a deterministic ARN from the (now shared) secret name, the API Gateway region, and the account, so it renders identically in both topologies and always matches one deterministic suppression.

- constants: add API_GATEWAY_AUTH_SECRET_NAME as the single source of truth for the secret name (previously hardcoded in two places).
- api_gateway_global_stack: create the secret from the constant.
- regional_stack: build the deterministic auth-secret ARN and pass api_gateway_region into the suppressions.
- nag_suppressions: scope the IAM5 auth-secret suppression to the API Gateway region (falls back to the global region).
- tests: add a single-region entry to the shared cdk config matrix so both the synthesis and cdk-nag matrices exercise the collapsed topology. This reproduces the exact IAM5 finding from the issue.

Fixes #125

## Type of change

## Testing

- [ ] `pytest tests/` passes locally
- [ ] `cdk synth` succeeds (if CDK code changed)
- [ ] New tests added for new behavior

## Checklist

- [ ] Documentation updated (README, `docs/`, inline docstrings) as needed
- [ ] No secrets, credentials, or customer data committed

### DIFF
--- a/gco/stacks/api_gateway_global_stack.py
+++ b/gco/stacks/api_gateway_global_stack.py
@@ -50,7 +50,7 @@ from aws_cdk import aws_secretsmanager as secretsmanager
 from aws_cdk import aws_wafv2 as wafv2
 from constructs import Construct
 
-from gco.stacks.constants import LAMBDA_PYTHON_RUNTIME
+from gco.stacks.constants import API_GATEWAY_AUTH_SECRET_NAME, LAMBDA_PYTHON_RUNTIME
 
 # <pyflowchart-code-diagram> BEGIN - auto-inserted, do not edit
 # Flowchart(s) generated from this file:
@@ -188,7 +188,7 @@ class GCOApiGatewayGlobalStack(Stack):
         secret = secretsmanager.Secret(
             self,
             "GCOAuthSecret",
-            secret_name="gco/api-gateway-auth-token",  # nosec B106 — this is the secret path, not a password
+            secret_name=API_GATEWAY_AUTH_SECRET_NAME,  # nosec B106 — this is the secret path, not a password
             description="Secret token for validating requests from API Gateway to ALB (auto-rotated)",
             generate_secret_string=secretsmanager.SecretStringGenerator(
                 secret_string_template=json.dumps({"description": "GCO API Gateway auth token"}),
--- a/gco/stacks/constants.py
+++ b/gco/stacks/constants.py
@@ -29,6 +29,34 @@ LAMBDA_PYTHON_RUNTIME = "PYTHON_3_14"
 """CDK enum name for the Lambda runtime (e.g. ``lambda_.Runtime.PYTHON_3_14``)."""
 
 # ---------------------------------------------------------------------------
+# API Gateway Auth Secret
+# ---------------------------------------------------------------------------
+# Physical name of the Secrets Manager secret that holds the token API Gateway
+# uses to authenticate requests to the regional ALBs. It is created by
+# ``GCOApiGatewayGlobalStack`` (in the ``api_gateway`` region) and read by the
+# regional service-account role and the regional API proxy Lambda.
+
+API_GATEWAY_AUTH_SECRET_NAME = "gco/api-gateway-auth-token"  # nosec B105 — secret path/name, not a credential
+"""Secrets Manager secret name for the API Gateway → ALB auth token.
+
+Single source of truth shared by three call sites that must agree exactly:
+
+1. ``GCOApiGatewayGlobalStack._create_secret`` — the ``secret_name`` the
+   secret is actually created with.
+2. ``GCORegionalStack`` — the deterministic IAM ``Resource`` ARN granting the
+   service-account role read access to the secret. Built from this name plus
+   the API Gateway region and account so it renders identically whether the
+   API Gateway stack is cross-region or co-located with the regional stack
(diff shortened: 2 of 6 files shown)


Jmevorach:
Closes https://github.com/awslabs/global-capacity-orchestrator-on-aws/issues/125

PR #150, opened 2026-07-13
Feature/job submission validation

JohnDzialo:
Add two hard-reject, configurable job-submission checks enforced identically by the REST manifest_processor and the SQS queue_processor so the SQS path is not a validation bypass:

- require_accelerator_toleration: jobs requesting nvidia.com/gpu, aws.amazon.com/neuron, or vpc.amazonaws.com/efa must carry a matching NoSchedule toleration, or they would stay Pending forever.
- max_parallelism: Job/CronJob spec.parallelism is capped.

Config lives under cdk.json::job_validation_policy and is plumbed to both services via {{MP_*}}/{{QP_*}} placeholders in regional_stack.py and the two kubectl-applier manifests.

## Type of change

- [ x] `feat:` New feature (non-breaking)

## Testing

- [x ] `pytest tests/` passes locally
- [ x] `cdk synth` succeeds (if CDK code changed)
- [ x] New tests added for new behavior

## Checklist

- [x ] Documentation updated (README, `docs/`, inline docstrings) as needed
- [ ] No secrets, credentials, or customer data committed


Jmevorach:
Thank you so much for the great PR; this is awesome @JohnDzialo!

Quick note on the CI fix I pushed here (5eca11a):

The only failing check was `lint:ruff:python` → `ruff format --check`. Ruff
wanted the parens dropped from two `except (TypeError, ValueError):` clauses
(`gco/services/manifest_processor.py`, `gco/services/queue_processor.py`).

It looks like the old Python 2 `except E, e:` syntax, but it isn't. This repo
targets Python 3.14, and ruff's formatter applies PEP 758
(https://peps.python.org/pep-0758/), which makes the parentheses around
multiple exception types optional when there's no `as` clause.
`except TypeError, ValueError:` is exactly equivalent to
`except (TypeError, ValueError):` — both catch either exception and no names
get rebound. Verified identical behavior on 3.14 before pushing.

All I did was run `ruff format`; nothing needed on your end. If we'd rather keep
the explicit parens for readability, that's a repo-level ruff config choice
(formatter target version), not a per-PR fix.

PR #153, opened 2026-07-14
fix: surface real AWS errors in capacity checks instead of masking as…

JohnDzialo:
… unavailable

The CLI capacity availability check caught ClientError/BotoCoreError and returned False ("assume unavailable"), so throttling, expired credentials, denied permissions, and non-opted-in regions were reported to users as "not available in region" — indistinguishable from genuine no-capacity, and the CLI exited 0 so the MCP wrapper reported success.

- Add CapacityCheckError (cli/capacity/models.py); raise it from check_instance_available_in_region on a failed offerings lookup. Only a genuinely-successful lookup returns True/False. estimate_capacity lets it propagate; the capacity command handlers' except-Exception turns it into a detailed non-zero exit, which cli_runner.py renders as an MCP {"error": ...}.
- Keep graceful degradation for secondary signals (spot score, AZs, AZ coverage, on-demand price) but log failures at warning so they're visible.
- multi_region.recommend_region_for_job now distinguishes "all region checks failed" (surfaces the underlying errors) from "no regions configured".
- Add CapacityEstimate.error field (default None) for explicit degraded-result errors; serializes via asdict into CLI/MCP JSON.
- Tests: CapacityCheckError propagation (RequestLimitExceeded/AuthFailure/ OptInRequired), no-regression on genuinely not-offered types, non-zero CLI exit with detailed message, and multi-region error surfacing.

## Type of change

- [ x] `fix:` Bug fix (non-breaking)

## Testing

- [x ] `pytest tests/` passes locally
- [x ] `cdk synth` succeeds (if CDK code changed)
- [ x] New tests added for new behavior

## Checklist

- [ ] Documentation updated (README, `docs/`, inline docstrings) as needed
- [x ] No secrets, credentials, or customer data committed

### DIFF
--- a/cli/capacity/__init__.py
+++ b/cli/capacity/__init__.py
@@ -29,6 +29,7 @@ from .history import (
 )
 from .models import (
     GPU_INSTANCE_SPECS,
+    CapacityCheckError,
     CapacityEstimate,
     InstanceTypeInfo,
     SpotPriceInfo,
@@ -45,6 +46,7 @@ __all__ = [
     "GPU_INSTANCE_SPECS",
     "BedrockCapacityAdvisor",
     "BedrockCapacityRecommendation",
+    "CapacityCheckError",
     "CapacityChecker",
     "CapacityEstimate",
     "CapacityHistoryStore",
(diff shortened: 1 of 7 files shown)


Jmevorach:
Thank you so much for the great PR; this is awesome @JohnDzialo! Merged and live on main in v3.14.2

PR #179, opened 2026-07-26
refactor: split the live-validation harness into reviewable modules

Jmevorach:
## Summary

The live-release-validation harness — GCO's end-to-end test — had grown into a 6,899-line `actions.py` holding 150 top-level symbols with no section markers, alongside a 1,487-line `inventory.py`. Finding the code behind a failing action meant scrolling, and the ownership/cleanup safety model was interleaved with the actions that consume it.

Both files are now packages organized by responsibility. Largest module is 856 lines (pre-existing `models.py`); largest action module is 203.

| Path | Responsibility |
|---|---|
| `actions/` | One module per registry entry — the "test case" layer |
| `checks/` | Reusable polling and validation helpers |
| `ownership/` | Durable proof of what this run created and may destroy |
| `cleanup/` | Deletion of exactly those proven-owned resources |
| `inventory/` | Read-only collection, split into `stacks`/`ecr`/`scanners`/`project` |
| `registry.py` | The ordered action registry, extracted from `actions.py` |
| `constants.py`, `context.py`, `protected.py` | Shared constants, run identity, ownership boundary |

**The move is behavior-neutral and mechanically verified.** All 205 symbols from `actions.py` and all 55 from `inventory.py` were relocated with byte-identical ASTs (compared symbol-by-symbol against `HEAD` before committing). The only genuinely new code is four functions from decomposing the worst offender, the 434-line `_cleanup_owned_log_groups`: its STS assume-role closure became `ownership.cleanup_role.TagConditionedLogDeleter`, its per-record body became `_converge_one_log_group`, plus a blocked-entry builder and a precondition validator. That function is now 293 lines and the safety-critical delete path is unchanged.

### Documentation

`scripts/live_release_validation/README.md` is the developer guide the harness never had. It answers the question that previously had no written answer — *where does my new check go?* — with an explicit decision table:

- a new assertion about infrastructure an existing action already inspects → extend that action
- a behavior needing no AWS account → a normal offline unit test, **not** here
- a new independently selectable phase with its own pass/fail meaning → a new action (4 mechanical steps, each enforced by a test)
- a new AWS resource type this run creates → an `ownership/` proof plus `cleanup/` deletion, following the "a matching resource name is never proof of ownership" rule

It also documents the layering rules and the testing workflow. `docs/LIVE_RELEASE_VALIDATION.md` (the operator runbook) now points at it.

### New tests

`tests/test_live_release_validation_structure.py` — 39 architecture guards so a structural mistake fails on the PR rather than two hours into a live run:

- every registry action has exactly one owning module and is exported from `actions/__init__.py`
- the registry agrees with the runbook contract table on action names, order, and dependencies
- dependencies are declared in an order the runner can actually execute
- layer imports flow one way, and only `runner.py`/`__main__.py` import `registry`
- no module exceeds a 900-line review ceiling (so a 6,900-line module cannot return)
- every module, action handler, and README guidance section stays documented

Two of these caught real problems while being written: a wrong layering assumption (`cleanup/` legitimately needs `checks/` job primitives) and `inventory.py` being over the ceiling, which is what prompted splitting it too rather than adding an exemption.

`tests/_live_validation_patching.py` — `patch_live_validation_helper` binds one shared mock into every module that references a name. Patching only the defining module would miss the binding the code under test actually calls, so a test could silently exercise the real implementation against a real account. This also means moving a helper between modules stays a pure refactor: no test needs a new patch target, and a stale target raises instead of passing.

## Type of change

- [x] `refactor:` Code refactor (no behavior change)
- [x] `test:` Test-only change

## Testing

- [x] `pytest tests/test_live_release_validation.py tests/test_live_release_validation_structure.py` — 142 passed
- [x] Symbol-level AST equivalence verified against `HEAD` for both split files (205 + 55 symbols, zero missing, zero changed, zero duplicated)
- [x] `ruff format --check .` and `ruff check .` clean; `mypy gco/ cli/ gco_mcp/ scripts/ --exclude 'gco/stacks/'` clean (218 files); `markdownlint` clean
- [x] `pytest tests/test_docs_coverage.py tests/test_doc_hygiene.py tests/test_bats_readme_coverage.py tests/test_integration.py` — 101 passed
- [ ] Live validation on this exact SHA — running after CI goes green; a sanitized `PASSED` summary will be posted as a comment

## Live release validation

- [x] Required — full `--actions all` run executed locally on `8002d6c` (this branch immediately before the final commit). The run itself surfaced a harness-internal manifest-path bug introduced by the modularization; it is fixed in the final commit (`8e874cf`) with offline regression tests that reproduced the live failure exactly. The maintainer reviewed the results and waived a fresh full run for the fix commit. Sanitized per-action summary posted as a PR comment.

**Applicability rationale:** Although this is a behavior-neutral refactor with no changes to deployed infrastructure, the code being restructured *is* the deploy/destroy/ownership harness itself. Offline tests cannot prove that the real teardown path still proves ownership correctly against live AWS, so the honest bar here is a full live run on this SHA.

## Checklist

- [x] Documentation updated (new developer README, runbook pointer, `tests/README.md` rows)
- [x] `requirements-lock.txt` regenerated if `pyproject.toml` changed (not applicable)

## Related issues

Follow-up candidates deliberately left out of scope to keep this diff mechanically verifiable: `destroy_deployment` (273 lines) and `_checkpoint_owned_log_groups` (324 lines) are long but are linear phase sequences; decomposing them warrants its own change with its own live run.


Jmevorach:
## Live release validation — sanitized summary

- **Run ID:** `retry1-8002d6c80f62`
- **Commit under test:** `8002d6c80f62cdcb6db3c2a2f5b48eaeed308722` (this branch, immediately before the final commit)
- **Scope:** `--actions all`, local operator run per `docs/LIVE_RELEASE_VALIDATION.md`; resumed twice with byte-identical `--resume` invocations after operator-side network interruptions

| Action | Status |
|---|---|
| `preflight` | passed |
| `baseline` | passed |
| `deploy` | passed |
| `topology` | passed |
| `api` | **failed** — harness-internal bug, see below |
| `sqs` | not reached |
| `central-queue` | not reached |
| `convergence` | not reached |
| `destroy` | passed |
| `final-inventory` | passed |

**Teardown/account state:** `final-inventory` passed with zero residual project resources in every deletable category (CloudFormation stacks, Global Accelerators, IAM, S3, regional resources). Remaining entries are the documented accepted classes only: 6 KMS keys inside their pending-deletion window and 2 intentionally retained ECR repositories.

**The `api` failure was a bug in the validation harness itself, not in GCO.** The modularization moved `_load_manifest` from the package root into `checks/jobs.py`; its `Path(__file__).with_name("manifests")` lookup was byte-identical but position-dependent, so it resolved `checks/manifests/` (which does not exist) and raised `FileNotFoundError` before submitting anything. Deploy, topology/routing, teardown, and ownership verification — the system-level paths this refactor was most likely to break — all executed against live AWS and passed.

**Fix:** `8e874cf` anchors the manifest directory at the package root (`_MANIFEST_DIR` in `constants.py`) and adds offline regression tests (`TestManifestPathResolution`) that drove the real loader path construction and reproduced the live failure exactly — same exception, same wrong path — before the fix. An audit of the package found this was the only position-sensitive construct (`__file__`/`__package__`/`inspect`/`importlib`) the refactor could have displaced.

**Maintainer decision:** system validated by this run; a fresh full live run for the fix commit is waived since the fix is harness-internal path resolution now covered by offline regression tests.

PR #184, opened 2026-07-28
Teardown hardening: sweep implicit orphans, retry inventory throttles

Jmevorach:
## Summary

Fixes the two teardown gaps surfaced by the last live release validation cycle:

### 1. `destroy-all` leaves implicit log groups + bastion IAM behind

A real teardown reported "All stacks destroyed successfully" while orphaning 22 CloudWatch log groups and the ephemeral bastion's IAM role + instance profile — which then failed the next validation run's clean-account `baseline` gate. CloudFormation only deletes the log groups it modeled; Lambda default groups (`/aws/lambda/<fn>`), the EKS control-plane group (`/aws/eks/<cluster>/cluster`), and the Container Insights groups are created out-of-band by the services.

Non-strict destroys now:
- capture the exact implicit log-group names derived from each stack's **own resources** (`list_stack_resources`) while the stacks still exist,
- delete those exact names only after their source stack's deletion succeeded (every exit path funnels through the sweep, so a partially failed teardown still cleans up the stacks that did delete),
- retire the bastion IAM role/profile via the existing validated teardown builders (`NoSuchEntity` reads as "nothing to clean").

Safety posture: deletion is by exact derived name only — never by prefix; explicit `AWS::Logs::LogGroup` resources stay CloudFormation-owned; both sweeps are best-effort and can never fail the destroy. **Strict (live-validation) teardowns skip both sweeps entirely** — the harness checkpoints, tags, and fences its own log-group generations and remains the single owner of that deletion authority (asserted by a dedicated test).

### 2. Harness inventory dies on transient throttling

`final-inventory` failed once with `ThrottlingException … reached max retries: 4` from CloudWatch Logs `ListTagsForResource` — the inventory issues one metadata read per resource across all 17 enabled Regions. The harness session is now a small facade (`scripts/live_release_validation/aws_session.py`) that gives every client botocore's `adaptive` retry mode: client-side rate limiting plus a deeper, still-bounded retry budget (12 attempts). Caller-supplied `Config`s keep every field they set; once the budget is exhausted the original `ClientError` still propagates, so all fail-closed paths behave exactly as before.

## Testing

- 19 new tests in `tests/test_stacks_destroy_hardening.py`: exact name derivation (Lambda/EKS/Container Insights; explicit log groups excluded), best-effort collection, deletion scoped to successfully destroyed stacks, missing-group tolerance, error swallowing, bastion IAM order/absence/failure handling, and `destroy_orchestrated` wiring — including the partial-failure sweep and the strict-mode "runs neither sweep" guarantee.
- 7 new tests in `tests/test_live_release_validation.py` for the session facade (adaptive config injection, merge semantics, delegation, runner wiring).
- `tests/conftest.py` extends the autouse no-op guard so orchestration tests can never fire the new AWS-backed helpers for real.
- Local runs, all green: `test_stacks_destroy_hardening.py` (54), `test_live_release_validation.py` (109), `test_stacks.py` (173), `test_stacks_extended.py` (116), `test_stacks_extended_coverage.py` (109), `test_cli_coverage.py` + `test_cli_command_gap_coverage.py` (57), `test_docs_coverage.py` + `test_doc_hygiene.py` (9).
- ruff format/check clean; mypy clean on the changed runtime files; markdownlint clean (110 files).
- Live E2E (full `scripts.live_release_validation` run) will be executed against this branch once CI is green; results will be posted before review.

## Docs

- `docs/CLI.md`: documents the new `destroy-all` sweep behavior.

Do not merge — for manual review per repo owner. The E2E result comment will follow.

### DIFF
--- a/cli/stacks.py
+++ b/cli/stacks.py
@@ -4214,6 +4214,13 @@ class StackManager:
         )
         record_cleanup("bastions", {"terminated_instances": bastions})
 
+        # Non-strict teardowns also retire the bastion's standing IAM
+        # role/profile and, below, the implicit log groups CloudFormation
+        # never modeled. Strict (live-validation) teardowns skip both: the
+        # harness owns fenced log-group deletion and audits IAM itself.
+        if not strict_identity:
+            record_cleanup("bastion-iam", self._cleanup_bastion_iam())
+
         global_stack_name = f"{project_name}-global"
         backup = self._cleanup_backup_vault(
             expected_stack_id=(expected_stack_ids or {}).get(global_stack_name),
@@ -4229,6 +4236,28 @@ class StackManager:
 
         successful: list[str] = []
         failed: list[str] = []
+
+        # Capture implicit log-group names while the source stacks still
+        # exist; the exact derived names are deleted by ``finish`` below
+        # once their stacks are gone. Strict teardowns collect nothing —
+        # the live-validation harness owns fenced log-group deletion.
+        implicit_log_groups: dict[str, dict[str, Any]] = {}
+        if not strict_identity:
+            implicit_log_groups = self._collect_implicit_log_groups(stacks)
+
+        def finish(overall: bool) -> tuple[bool, list[str], list[str]]:
+            """Funnel every exit through the implicit log-group sweep.
+
+            Called at each return point so a partially failed teardown
+            still cleans up the stacks that DID delete. New exit paths
+            must return through here as well.
+            """
+            if implicit_log_groups:
+                record_cleanup(
+                    "implicit-log-groups",
+                    self._cleanup_implicit_log_groups(implicit_log_groups, successful),
+                )
+            return overall, successful, failed
 
         for stack_name in post_regional_stacks:
             if on_stack_start:
@@ -4252,7 +4281,7 @@ class StackManager:
             if stack_name not in failed:
                 failed.append(stack_name)
         if any(stack in failed for stack in post_regional_stacks) or phase_remaining:
-            return False, successful, failed
+            return finish(False)
 
         if regional_api_stacks:
             if parallel and len(regional_api_stacks) > 1:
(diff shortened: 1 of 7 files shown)


Jmevorach:

Full E2E run against this branch's head completed with every action green on the first attempt — no resume needed.

- Run ID: `local-20260728T012248Z-7c43b4f15922`
- SHA: `7c43b4f1592294f9d425cc138dcff9dc859bccc6` (this PR's head)
- Overall status: **passed**, `fatal_error: none`

|---|---|---|
| preflight | passed | 103s |
| baseline | passed | 8m 50s |
| deploy (5 stacks) | passed | 47m 34s |
| topology | passed | 8m 22s |
| api | passed | 26s |
| sqs | passed | 45s |
| central-queue | passed | 30s |
| opencost | passed | 1s |
| convergence | passed | 32s |
| destroy (5 stacks) | passed | 82m 22s |
| final-inventory | passed | 8m 44s |

Final inventory: zero residual project resources, protected baseline (`CDKToolkit`, `GCOGitHubOIDCStack`) byte-identical, expected accepted residuals only (KMS keys in their 7-day PendingDeletion window + CDK container-asset ECR digests).

### Evidence the two fixes behaved

- **Throttle resilience:** the previous run on `main` failed `final-inventory` with `ThrottlingException … reached max retries: 4` and needed an internal retry + resume. This run's `final-inventory` (same 17-Region, per-resource tag scan) passed cleanly on the first attempt through the new adaptive-retry session.
- **Strict-mode ownership preserved:** the harness's own fenced destroy + log-group cleanup ran exactly as before (the new destroy sweeps are non-strict-only by design), and the post-run account is verifiably clean — zero gco stacks, zero gco log groups, no bastion IAM role/profile in the account (checked directly, not just via the report).

The non-strict `destroy-all` sweep itself is covered by the 19 new unit tests in this PR (the E2E exercises the strict path, which must — and did — stay byte-identical).

Full reports remain local per the validation doctrine. Ready for review; not merging anything myself.

PR #213, opened 2026-08-01
fix: honor structured capacity output

### LINKED ISSUE #212: [bug] gco capacity subcommands ignore --output json and emit human-formatted text
# `gco capacity` subcommands ignore `--output json` and emit human-formatted text

**Version:** 5.2.2 (git `491b84a`)

## Summary

Three `gco capacity` subcommands ignore the global `--output json` flag and print
human-formatted text (box-drawn tables, `✓`/`ℹ` glyphs) instead of JSON. Any
programmatic caller receives unparseable output — including **GCO's own MCP
server**, which hardcodes `--output json` for every CLI invocation
(`gco_mcp/cli_runner.py:104`).

## Affected commands

| Command | `--output json` respected? | First line of output |
|---|---|---|
| `capacity check -i <type> -r <region>` | ✅ yes | `[` |
| `capacity instance-info <type>` | ✅ yes | `{` |
| `capacity spot-prices -i <type> -r <region>` | ✅ yes | `[` |
| `capacity status` *(all regions)* | ❌ **no** | `  REGION  QUEUE  RUNNING  GPU%…` |
| `capacity status -r <region>` | ✅ yes | `{` |
| `capacity recommend-region [--gpu] [-i <type>]` | ❌ **no** | `✓ Recommended region: us-east-1` |
| `capacity recommend -i <type> -r <region>` | ❌ **no** | `ℹ Recommended: ON-DEMAND` |

Note the asymmetry within a single command: `capacity status -r us-east-1` returns
clean JSON, but the all-regions form does not.

## Steps to reproduce

```bash
# ✅ Correct behavior (JSON):
gco --output json capacity check -i g5.2xlarge -r us-east-1
gco --output json capacity status -r us-east-1

# ❌ Bug — all three print human text despite --output json:
gco --output json capacity status
gco --output json capacity recommend-region --gpu -i g5.2xlarge
gco --output json capacity recommend -i g5.2xlarge -r us-east-1
```

**Actual** output of `gco --output json capacity status`:

```
  REGION          QUEUE  RUNNING  GPU%   CPU%   SCORE
  -------------------------------------------------------
  us-east-1           0        0     0%     0%   2000

ℹ Recommended region: us-east-1 (lowest score = best)
```

**Expected:** a JSON document — e.g. an array of per-region capacity objects,
matching the shape `capacity status -r <region>` already returns.

## Root cause

In `cli/commands/capacity_cmd.py`, the working commands branch on the requested
format and hand the **data** to the formatter; the broken ones call human-only
helpers unconditionally.

**Working pattern** (`capacity check`, ~line 53):

```python
formatter = get_output_formatter(config)
if config.output_format == "table":
    print(format_capacity_table(estimates))
else:
    formatter.print(estimates)          # honors json/yaml
```

**Broken pattern 1 — no format branch at all** (`recommend-region`, ~line 235;
same shape in `recommend`, ~line 85):

```python
formatter = get_output_formatter(config)
recommendation = checker.recommend_region_for_job(...)
formatter.print_success(f"Recommended region: {recommendation['region']}")
formatter.print_info(f"Reason: {recommendation['reason']}")
```

The `recommendation` dict is already structured — it is simply never printed as
data.

**Broken pattern 2 — JSON emitted, then polluted** (`capacity status`, ~line 172).
The single-region branch is correct, but the all-regions branch hardcodes a table
and appends an `ℹ` line regardless of format:

```python
if region:
    capacity = checker.get_region_capacity(region)
    formatter.print(capacity)                    # ✅ correct
else:
    capacities = checker.get_all_regions_capacity()
    print("\n  REGION          QUEUE  RUNNING  GPU%   CPU%   SCORE")   # ❌ always
    ...
    formatter.print_info(f"Recommended region: {best.region} …")       # ❌ always
```

Underlying reason these slip through: in `cli/output.py`, `print_success()`,
`print_info()`, `print_warning()` and `print_error()` are **format-agnostic** —
they unconditionally emit `✓ …` / `ℹ …` regardless of `output_format`:

```python
def print_success(self, message: str) -> None:
    print(f"✓ {message}")

def print_info(self, message: str) -> None:
    print(f"ℹ {message}")
```

So any command whose only output path is those helpers can never produce JSON.

## Impact

1. **GCO's own MCP server is affected.** `gco_mcp/cli_runner.py` builds every
   command as `[gco, "--output", "json", *args]`, so the MCP tools
   `capacity_status`, `recommend_region` and `recommend_capacity` return
   decorated text to their callers. An LLM client may cope; deterministic code
   cannot.
2. **The purpose-built placement tool can't be consumed programmatically.**
   `recommend-region` carries GCO's own multi-signal scoring (spot placement
   scores, queue depth, pricing) — exactly what an external scheduler or
   orchestrator wants — and it's the one that can't be parsed.
3. **Silent-failure risk for consumers.** A caller that `json.loads()` this gets a
   clean exception, which is fine. But `capacity status`'s "JSON then text"
   behavior is more dangerous: a lenient parser can succeed on the leading
   fragment and silently miss the rest, so a consumer could mis-rank regions and
   place workloads on incomplete signal.

## Workaround for consumers

Route around the affected commands rather than scraping their output: the
JSON-returning `capacity check` carries per-AZ `availability`, `confidence`,
`price_per_hour` and `details.on_demand_price`, which can be collapsed into a
per-region verdict client-side. `capacity status -r <region>` (one call per
region) also returns clean JSON. Text responses should be rejected outright
rather than partially parsed, so a decorated table can't masquerade as a capacity
signal.

## Suggested fix

1. In `capacity_cmd.py`, give `recommend-region`, `recommend`, and the all-regions
   branch of `status` the same format branch the working commands use — pass the
   existing structured dict/list to `formatter.print(...)` when
   `config.output_format != "table"`, keeping the human rendering for `table`.
2. Consider making the `print_*` helpers in `cli/output.py` no-ops, or routing
   them to **stderr**, when `output_format` is `json`/`yaml`. That turns this
   class of bug into a non-issue repo-wide rather than patching three call sites:
   any command that only calls the human helpers would then emit nothing on
   stdout instead of invalid JSON. Sending human chatter to stderr also keeps
   `--output json` stdout machine-clean while a human still sees progress.
3. Worth sweeping the other command groups (`jobs`, `stacks`, `images`, `costs`,
   `inference`) for the same pattern — only `capacity` was audited here.

## Verification

After the fix, all of these should emit parseable JSON with no glyphs (this
snippet fails on 5.2.2 today):

```bash
for c in "capacity status" \
         "capacity recommend-region --gpu -i g5.2xlarge" \
         "capacity recommend -i g5.2xlarge -r us-east-1"; do
  gco --output json $c | python3 -c 'import json,sys; json.load(sys.stdin); print("OK")'
done
```

## Environment

- GCO `5.2.2`, git `491b84a`
- macOS (darwin 25.5.0), Python 3.14 venv in the GCO checkout
- Region `us-east-1`
- Note: the test account had **no GCO stack deployed** (`gco-api-gateway`
  missing), so `capacity status` reported a single region with zeros. The
  formatting bug is independent of deployment state — it reproduces on the code
  path, not the data.

Jmevorach:
## Summary

- route all-region capacity status through the configured formatter for JSON and YAML output
- return structured capacity recommendations instead of success/info text outside table mode
- emit an empty structured result when no GCO stacks are deployed while retaining the warning on stderr
- keep existing human-readable table output, including verbose region rankings
- add regression coverage for all affected JSON paths and table compatibility

## Type of change

## Testing

- [ ] `pytest tests/` passes locally
- [ ] `cdk synth` succeeds (if CDK code changed)

Validation completed:
- `pytest tests/test_capacity.py tests/test_capacity_cmd_coverage.py tests/test_cli_command_gap_coverage.py -q` (217 passed)
- `ruff format --check cli/commands/capacity_cmd.py tests/test_cli_command_gap_coverage.py`
- `ruff check cli/commands/capacity_cmd.py tests/test_cli_command_gap_coverage.py`
- actual workspace-backed `gco` 5.3.3 commands were piped through `json.load`: `capacity status`, `capacity recommend-region --gpu -i g5.2xlarge`, and `capacity recommend -i g5.2xlarge -r us-east-1` all parsed successfully
- the local account has no deployed GCO stacks, confirming `capacity status` emits `[]` on stdout and keeps its warning on stderr

The documented mypy command is currently blocked by unrelated existing CDK typing errors in `gco/stacks`, `app.py`, and `tests/test_nag_compliance.py`.

## Live release validation

- [x] Not required (explain the applicability decision below)
- [ ] Required and completed locally with `--actions all` for this exact SHA; a sanitized `PASSED` summary comment (run ID, SHA, per-action statuses) was posted on the pull request
- [ ] Required but pending explicit validation-account and KMS-deletion authorization

**Applicability rationale:** This is an isolated CLI rendering fix exercised with mocked tests and read-only local CLI checks. It does not change infrastructure, AWS resource mutation, IAM, networking, or deployed runtime behavior.

**Sanitized validation summary or comment link:** Not applicable.

## Checklist

- [x] Documentation updated (README, `docs/`, inline docstrings) as needed — no documentation changes are needed because this restores the documented output contract
- [x] `requirements-lock.txt` regenerated if `pyproject.toml` changed — not applicable

## Related issues

Fixes #212

### DIFF
--- a/cli/commands/capacity_cmd.py
+++ b/cli/commands/capacity_cmd.py
@@ -190,6 +190,8 @@ def capacity_status(config: Any, region: Any, all_regions: Any) -> None:
 
             if not capacities:
                 formatter.print_warning("No GCO stacks found")
+                if config.output_format != "table":
+                    formatter.print(capacities)
                 return
 
             if config.output_format == "table":
--- a/cli/commands/capacity_cmd.py
+++ b/cli/commands/capacity_cmd.py
@@ -90,8 +90,16 @@ def recommend_capacity(config: Any, instance_type: Any, region: Any, fault_toler
             instance_type, region, fault_tolerance
         )
 
-        formatter.print_info(f"Recommended: {capacity_type.upper()}")
-        formatter.print_info(f"Reason: {explanation}")
+        if config.output_format == "table":
+            formatter.print_info(f"Recommended: {capacity_type.upper()}")
+            formatter.print_info(f"Reason: {explanation}")
+        else:
+            formatter.print(
+                {
+                    "capacity_type": capacity_type,
+                    "explanation": explanation,
+                }
+            )
 
     except Exception as e:
         formatter.print_error(f"Failed to get recommendation: {e}")
@@ -184,19 +192,21 @@ def capacity_status(config: Any, region: Any, all_regions: Any) -> None:
                 formatter.print_warning("No GCO stacks found")
                 return
 
-            # Format as table
-            print("\n  REGION          QUEUE  RUNNING  GPU%   CPU%   SCORE")
-            print("  " + "-" * 55)
-            for c in sorted(capacities, key=lambda x: x.recommendation_score):
-                print(
-                    f"  {c.region:<15} {c.queue_depth:>5}  {c.running_jobs:>7}  "
-                    f"{c.gpu_utilization:>4.0f}%  {c.cpu_utilization:>4.0f}%  {c.recommendation_score:>5.0f}"
-                )
+            if config.output_format == "table":
+                print("\n  REGION          QUEUE  RUNNING  GPU%   CPU%   SCORE")
+                print("  " + "-" * 55)
+                for c in sorted(capacities, key=lambda x: x.recommendation_score):
+                    print(
+                        f"  {c.region:<15} {c.queue_depth:>5}  {c.running_jobs:>7}  "
+                        f"{c.gpu_utilization:>4.0f}%  {c.cpu_utilization:>4.0f}%  "
+                        f"{c.recommendation_score:>5.0f}"
+                    )
 
-            # Show recommendation
-            print()
-            best = min(capacities, key=lambda x: x.recommendation_score)
-            formatter.print_info(f"Recommended region: {best.region} (lowest score = best)")
+                print()
+                best = min(capacities, key=lambda x: x.recommendation_score)
+                formatter.print_info(f"Recommended region: {best.region} (lowest score = best)")
+            else:
+                formatter.print(capacities)
 
     except Exception as e:
         formatter.print_error(f"Failed to get capacity status: {e}")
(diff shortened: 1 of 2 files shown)


Jmevorach:
Closes: https://github.com/awslabs/global-capacity-orchestrator-on-aws/issues/212

PR #224, opened 2026-08-02
feat: managed deployment-config engine + gated region tools

### LINKED ISSUE #221: [feat] `add_deployment_region` — manage deployment regions without editing `cdk.json`
# Feature request: `add_deployment_region` — manage deployment regions without editing `cdk.json`

**Version observed:** 5.2.2 (git `491b84a`)

## Problem

Adding a new regional deployment target requires a human to clone the GCO
repository and hand-edit `cdk.json`:

```json
"deployment_regions": {
  "global": "us-east-2",
  "api_gateway": "us-east-2",
  "monitoring": "us-east-2",
  "regional": ["us-east-1"]        // <- must be edited by hand to add a region
}
```

There is no CLI command or MCP tool for it. `gco config-cmd` is read-only
(`get`/`init`/`show`), and `gco stacks` has no region-management subcommand, so
the only path is a text edit inside a checkout.

This blocks three things:

1. **Programmatic/agent-driven use.** A client driving GCO over MCP can provision
   a region end-to-end (`deploy_stack`, `destroy_stack`, image publish, job
   submit) but cannot *add* one. The automation chain breaks at a text edit, so
   an otherwise fully-automatable workflow needs a human to stop and edit a file.

2. **Hosted / remote deployments.** When GCO's MCP server runs somewhere other
   than the operator's laptop, there is no local checkout to edit. The person who
   wants a new region may have no filesystem access to the repo at all.

3. **Auditability.** A hand-edit leaves no record — no actor, no timestamp, no
   approval trail — for a change that determines where infrastructure gets
   created and billed. Every other consequential GCO action is an auditable tool
   call (`@audit_logged`); this one is invisible.

Worth noting the shape of the gap: `global`, `api_gateway` and `monitoring` are
single-region scalars, while `regional` is described in the config's own comment
as *"the unlimited list of EKS clusters"* — the field most expected to change
over time is the one with no management interface.

## Proposed solution

A small set of region-management MCP tools (with CLI equivalents) that mutate
`deployment_regions` server-side and re-synthesize, mirroring the pattern GCO
already uses for other `cdk.json` flips:

```
add_deployment_region(region: str, role: str = "regional") -> str
remove_deployment_region(region: str, role: str = "regional") -> str
list_deployment_regions() -> str
```

CLI equivalents:

```bash
gco stacks regions list
gco stacks regions add us-east-2
gco stacks regions remove us-east-2
gco --output json stacks regions list      # machine-readable
```

Suggested behavior:

- **Validate before writing** — reject a region that is not SDK-known, or that
  belongs to a different AWS partition than the already-configured regions
  (`cdk.json`'s own comment states both constraints, so the rules already exist
  and are currently enforced only implicitly at synth time).
- **Idempotent** — adding an already-present region succeeds without duplicating
  it; removing an absent one is a no-op with a clear message.
- **Report what changed and what it implies** — e.g. `added us-east-2 to
  'regional'; new stacks available: gco-us-east-2, gco-regional-api-us-east-2.
  Run 'gco stacks deploy gco-us-east-2' to provision.`
- **Do not deploy as a side effect.** Adding a region should only change
  configuration; provisioning stays an explicit, separately-gated action. This
  keeps the destructive/costly step where operators already expect it.
- **Removal should not silently orphan infrastructure.** If a deployed stack
  exists for the region being removed, warn (or require a `--force`), since
  dropping it from config makes the running stack invisible to `stacks list`
  while it continues to bill.
- **Gate it** behind `GCO_ENABLE_INFRASTRUCTURE_DEPLOY` (or a dedicated flag) —
  it changes what infrastructure can be created, so it belongs with the other
  deploy-class tools rather than the read-only set.
- **Audit it** with `@audit_logged`, like the rest of the mutating surface.

There is already an in-repo precedent for a tool that edits `cdk.json`:
`enable_fsx` / `disable_fsx` are documented as *"flip FSx Lustre on/off in
cdk.json"* and are exposed as MCP tools (`gco stacks fsx enable|disable`). Region
management is the odd one out; the mechanism it needs already exists in the same
module.

## Alternatives considered

- **CDK context override at deploy time** (`cdk deploy -c
  deployment_regions=...`). Rejected: `gco stacks deploy` exposes no context
  passthrough, and even if added, a per-invocation override would not persist, so
  `stacks list` / `status` would disagree with what was actually deployed.

- **An environment variable** (e.g. `GCO_REGIONAL_REGIONS=us-east-1,us-east-2`).
  Simple, and works for hosted deployments, but splits configuration across two
  sources of truth. Reasonable as a supplement; not a substitute for a managed,
  auditable mutation.

- **External clients editing `cdk.json` themselves.** Rejected on principle: it
  couples every consumer to GCO's internal file layout and JSON schema, so a
  refactor inside GCO silently breaks them. Config ownership should stay with
  GCO, exposed through an interface.

- **Documenting the manual edit more clearly.** Helps a human operator, but does
  nothing for programmatic callers or hosted deployments, and leaves the
  audit-trail gap untouched.

## Scope

- **CLI change** — new `gco stacks regions {list,add,remove}` subcommands,
  honoring `--output json` (see the related output-format issue below).
- **MCP tool addition** — `add_deployment_region` / `remove_deployment_region` /
  `list_deployment_regions` in `gco_mcp/tools/stacks.py`, registered in
  `run_mcp.py` and added to the tool gating table.
- **Config-layer change** — validated read/modify/write of
  `deployment_regions` in `cdk.json` (partition and SDK-known-region checks).
- **Docs** — update the region-management section to describe the tool as the
  supported path instead of a manual edit.
- No CDK stack change required: the synthesized stacks already derive from
  `deployment_regions` (verified — adding `us-east-2` by hand produced
  `gco-us-east-2` and `gco-regional-api-us-east-2` on the next synth).

## Additional context

- Verified on 5.2.2 (`491b84a`): after hand-adding `us-east-2` to `regional`,
  `gco stacks list` reported the new `gco-us-east-2` and
  `gco-regional-api-us-east-2` stacks, and `gco stacks deploy gco-us-east-2`
  provisioned successfully. The synthesis side already works — only the
  configuration interface is missing.
- Related: `capacity status`, `recommend-region` and `recommend` ignore
  `--output json` and emit human-formatted text. Any new `stacks regions`
  subcommand should honor the flag from the start so programmatic callers can
  consume it.
- Prior art in this repo: `enable_fsx` / `disable_fsx` (`gco stacks fsx
  enable|disable`) already mutate `cdk.json` through a tool, plus the
  Aurora/Valkey config commands under `gco stacks`.

Jmevorach:
Closes #221.

## Summary

Instead of adding one more bespoke toggle for the regions knob, this PR introduces a small **managed deployment-config engine** and ships every deployment-config knob from the #221 discussion as registered keys: the workload-Region list, the three control-plane Region scalars, and the Bedrock model default.

## The engine (`cli/managed_config.py`)

Each externally manageable cdk.json key registers a `ManagedListKey` (set-semantics list) or `ManagedScalarKey` (single string). Every mutation flows through one core that guarantees:

- **Resolution** — targets the same `cdk.json` the CDK CLI would use (cwd upward), or an explicit `--config-path`. Installed (`uvx`/`pip`) distributions resolve read-only package data and are **refused with guidance** instead of half-working.
- **Result-only validation** — an edit is accepted iff the *resulting* config passes the exact synth-time validators (`gco/stacks/constants.py`: SDK-known Regions, unique, non-empty, one AWS partition; for the Bedrock knob, the runtime reader contract from `gco/bedrock.py`). Validating the result rather than the starting state deliberately keeps the **repair path** open: removing a typo'd Region from a hand-edited config works.
- **Idempotency** — re-adding a present value, removing an absent one, or setting the current value is a reported no-op; zero bytes written.
- **Atomicity** — tmp + `os.replace` like the existing feature toggles, preserving `_comment_*` keys, key order, sibling settings (e.g. `bedrock.thinking`), file mode, and the original trailing-newline state.
- **Auditability** — a uniform `ChangeReport` plus structured log lines on `gco.cli.managed_config`; the MCP surface adds `@audit_logged` on top.

## Veneers

- **CLI**: `gco stacks regions list/add/remove/set` and `gco stacks bedrock show/set-model`. `list` shows the topology, resolved partition, and backing path — and diagnoses broken configs via `partition_error`. Mutating commands confirm unless `-y`, never deploy, and say so (`remove`/`set` explicitly warn that deployed stacks are not destroyed or moved).
- **MCP**: `list_deployment_regions` / `add_deployment_region` / `remove_deployment_region` / `set_deployment_region` / `set_default_bedrock_model`, registered only under the new default-off **`GCO_ENABLE_CONFIG_MANAGEMENT`** flag, shelling to the CLI like every other gated tool. Flag wired through `feature_flags.py`, the autopilot flag mirror, the `run_mcp` reload-compat block, and the `mcp://gco/feature-flags` gating table. Tool counts: 134 default, ceiling 174 → **179**.

## Tests

`tests/test_managed_config.py` — 68 tests: engine resolution / validation / idempotency / write mechanics (comments, order, mode, newline, siblings) / repair / status / audit lines for both key kinds, CliRunner coverage for all eight commands, MCP gating + argv translation. Guard updates: tool-count and name-set guards, `UMBRELLA_FLAG_TOOLS`, the feature-flags resource map, doc-coverage and tool-count-docs guards (all green).

## Ride-alongs (requested)

- **CI**: `unit:pytest:core` (the shard-coverage combine job) now installs only the `coverage` pin greped from `requirements-lock.txt` instead of the full editable `[dev]` extra — the job never imports the project. Verified locally that a coverage-only venv at the locked version combines real shard data files written by a full environment and produces every report format.
- **README**: escaped the single tildes in the GPU cost estimate line that GitHub rendered as strikethrough.

## Testing

- `pytest`: targeted suite (managed-config + MCP server/gating/flags/self-resources/autopilot + doc guards) — 403 passed; the one failure (`test_models_upload_argv_minimal`) reproduces on pristine main (local macOS `/dev/fd` quirk) and is unrelated.
- `ruff check` + `ruff format --check` (repo-wide), `mypy` on touched files: clean.
- `yamllint` + `actionlint` on the workflow change; `markdownlint-cli2` on all touched docs: clean.


Jmevorach:
Closes: https://github.com/awslabs/global-capacity-orchestrator-on-aws/issues/221

PR #245, opened 2026-08-13
Mission memory: recall lessons from similar past missions

Jmevorach:
## Summary

Missions now leave institutional memory behind. Every session that reaches a terminal verdict writes one memory item — the verbatim directive, its embedding, the report's lessons and recommended follow-ups, the verdict, and embedding provenance — to a new `{project}-mission-memory` DynamoDB table, and future sampling sessions retrieve the closest past missions by directive similarity into the strategy-revision prompt.

Phase 1 of the vector-search plan; built on DynamoDB's native vector indexes.

### Infrastructure
- `mission_memory` cdk.json block (**enabled by default** — one small pay-per-request item plus one embedding call per completed mission; silently missing recall is the worse failure mode; `enabled: false` opts out).
- Global stack provisions the table (PAY_PER_REQUEST, PITR, TTL, AWS-managed encryption, backup-plan membership) and the `directive-embedding-index` vector index via an `AwsCustomResource` — CloudFormation cannot express vector indexes natively. The custom resource runs under a dedicated pre-created role scoped to the exact table ARN. Table/index names publish to SSM for runtime discovery.
- The disabled configuration synthesises a **byte-identical** template (asserted in the synthesis matrix), so opting out is provably zero-footprint.

### Runtime
- `gco_mcp/mission/embeddings.py` — directive embedding through the configured Bedrock model (`context.bedrock.embedding_model_id`, stock `amazon.titan-embed-text-v2:0`), typed errors throughout.
- `gco_mcp/mission/memory.py` — `MissionMemoryStore` (write / similarity search / list) with SSM-lazy name resolution and a strict error taxonomy: absent table, backfilling index, missing SSM parameters, and missing credentials all surface as "unavailable" so callers degrade cleanly.
- Engine wiring is **best-effort everywhere**: the terminal write reuses the report's lessons overlay (never re-samples) and swallows every failure; retrieval is gated on `use_sampling` so the deterministic Propose path stays free of network calls — the determinism suite passes unchanged. The prior-missions prompt block has its own 4 KiB truncation domain and its absence keeps prompts byte-identical to today's.

### Surfaces
- MCP: `mission_memory_search` (tenth `mission_*` tool, behind `GCO_ENABLE_MISSION`).
- CLI: `gco mission memory search|list|backfill` — backfill seeds memory from existing `*.report.json` Final_Reports so recall is useful on day one.

### Docs
MISSION.md "Mission Memory" section (including the three one-way doors: dimensions, distance function, projection), CUSTOMIZATION.md embedding-model key, CLI.md command reference, module/tools/tests READMEs, regenerated code diagrams.

## Testing
- 5 new test files (config, stack synthesis, runtime store/embeddings with mocked boto3, engine fault-injection + prompt block + factory retrieval, CLI) — ~130 new tests, no live AWS anywhere.
- Fault injection pins the contract: absent table, backfilling index, Bedrock down, and store bugs never fail a mission.
- `test_mission_decide_determinism.py` and the full mission e2e corpus pass unchanged; cdk-nag zero unsuppressed findings; both synth states clean; docs-coverage gates green.
- A suite-wide conftest neutraliser patches the store-construction seam so no test can implicitly write to a real institutional-memory table from a credentialed developer machine.

## Notes for reviewers
- `SearchVectors` request-shape gotcha is pinned in code and tests: `SearchVector` is a plain list of `{"N": ...}` values — the `L` wrapper is only the item-attribute write shape.
- `dimensions` / `distance_function` / the `INCLUDE` projection are immutable after index creation; defaults (1024 / COSINE) match Titan V2 and are enforced against cdk.json by consistency tests.

This work was directly inspired by a respected technical voice in the company whose exploration of vector-index-backed agent memory shaped the design here — thank you, Eliad Maimon.


Jmevorach:
## Live verification on account infrastructure (us-east-2)

Deployed `gco-global` from this branch and exercised every leg of the feature against real AWS:

- **Provisioning**: stack CREATE_COMPLETE; `gco-mission-memory` table ACTIVE; `directive-embedding-index` created by the custom resource with the exact designed shape (1024-dim COSINE, `final_verdict` INLINE_FILTER, 7-attribute INCLUDE projection); both SSM parameters published.
- **Backfilling degradation**: a search issued while the index was still backfilling produced the friendly deployment hint + structured envelope + exit 1 (service error: "Cannot search backfilling vector index") — exactly the designed unavailable path.
- **Day-one backfill**: `gco mission memory backfill` embedded and wrote all 26 pre-existing Final_Reports (`{"written": 26, "skipped": 0, "failed": 0}`) — first live exercise of the Titan V2 embedding path.
- **Search quality**: near-verbatim directives match at COSINE distance ~0.02; semantically related queries order sensibly; `--verdict complete` inline filter returns only completed missions.
- **Engine write + recall**: a fresh `gco mission run` (Bedrock-scaffolded criteria, complete in 1 iteration) automatically left a memory item on terminal verdict, retrievable as the top hit (distance 0.066) for a paraphrased query.

### Three deploy-time findings, each fixed + pinned by a regression test
1. **Bundled Lambda SDK predates vector indexes** — AwsCustomResource silently dropped `VectorIndexUpdates` at serialization. Fixed with `install_latest_aws_sdk=True` (deploy-time only; bounded risk documented in-code with a revisit note).
2. **A failed create wedged rollback in DELETE_FAILED** — the delete path failed identically. Fixed with `ignore_error_codes_matching` on on-delete; proven live twice (both subsequent failed deploys rolled back clean).
3. **`CreateVectorIndexAction` is flat** (not the nested draft shape) **and the SearchSchema attribute needs an `AttributeDefinitions` entry in the same UpdateTable call** (the add-a-GSI pattern). Both fixed; the synthesized Create/Delete payloads are now walked recursively against botocore's UpdateTable model so any unknown or missing member fails at unit-test time instead of deploy time.

All three gotchas are also recorded in the Phase 1 spec for Phase 2 to inherit. CI green (79 checks) on the final commit.

PR #277, opened 2026-08-19
feat: capacity-driven Global Accelerator traffic dialing

Jmevorach:
## Summary

Adds capacity-driven Global Accelerator traffic dialing: an opt-in, EventBridge-scheduled controller Lambda in the global stack that converges each region's endpoint-group `TrafficDialPercentage` toward that cluster's `GCO/HealthMonitor` `ClusterHealthy` signal, so a region with degraded GPU pools sheds new connections gradually instead of waiting for binary health-check failover. Inspired by the Global Accelerator support announcement in the AWS Load Balancer Controller (the CRD's ownership model doesn't fit GCO's one-accelerator/N-clusters topology, but `trafficDialPercentage` was the missing capability).

- `global_accelerator.traffic_dial` cdk.json block (default **disabled**; `monitor` mode computes and publishes decisions only, `enforce` applies them and is the only shape granted `globalaccelerator:UpdateEndpointGroup`)
- Safety rails: dial floor, per-run step limit, last-healthy-region guard (never leaves every group below 100), hold on missing telemetry, skip while the accelerator is mid-deployment, per-region operator overrides
- New CLI: `gco capacity traffic-dial show|set|clear` (manual dial + override lifecycle over the same SSM tree the controller honors)
- Decisions published to `GCO/TrafficDial` CloudWatch metrics and `/{project}/traffic-dial/state` in SSM

Also fixes the GA health-check contract while in the area:

- Removes the never-consumed `health_check_grace_period` / `health_check_timeout` knobs (Global Accelerator has no such settings; legacy configs carrying them still synthesize), adds `health_check_threshold`, restricts `health_check_interval` to the API's only accepted values `{10, 30}`
- `get_global_accelerator_config` now merges partial blocks over defaults instead of the historical all-or-nothing fallback
- `ensure_https_health_check` no longer silently resets a configured interval to 30: the configured contract rides the convergence payload as synth-time literals (replayed pre-upgrade `gco stacks addons install` inputs keep working via handler defaults)

Hardening: `seccompProfile: RuntimeDefault` on the six platform pod specs (the only securityContext field they were missing).

Also folded in while driving this branch through live validation:

- Issue #278 dependency updates across every pin surface (Python + lockfile, npm, Dockerfile.dev toolchain, Trivy, Helm charts, EKS add-ons, container images), with a suppression sweep that kept all 25 `.trivyignore` entries on evidence
[...]

PR #281, opened 2026-08-20
feat: cleanup dynamic EBS volumes on regional stack destroy

### LINKED ISSUE #268: [feat] `gco stacks destroy` leaves the cluster's dynamically-provisioned EBS volumes behind
# GCO: `stacks destroy` leaves dynamically-provisioned EBS volumes behind

**Type:** GCO feature request / cleanup gap  
**Severity:** medium — small per occurrence, unbounded across repeated deploy/destroy cycles
**Found:** 2026-08-17 and 2026-08-18, two consecutive regional teardowns in `us-west-2`

---

### Summary

After `gco stacks destroy gco-<region>` completes successfully — stack gone, EKS
cluster gone, NAT gateways gone, VPC gone — the EBS volumes that the cluster's
EBS CSI driver provisioned for in-cluster PVCs **remain**, in `available` state,
tagged for a cluster that no longer exists. Nothing can ever reattach them, and
they keep billing.

Deleting an EKS cluster does not delete the PersistentVolumes its CSI driver
created, so any chart with a PVC leaves storage behind. With the shipped
defaults that is the observability stack: `cluster_observability` is **enabled by
default**, and kube-prometheus-stack provisions PVCs for both Prometheus and
Alertmanager.

### Reproduction

```bash
gco stacks deploy gco-<region> -y        # cluster_observability on by default
gco stacks destroy gco-<region> -y       # reports success; stack/EKS/NAT all gone

# Volumes tagged for the now-deleted cluster survive:
aws ec2 describe-volumes --region <region> \
  --filters Name=tag-key,Values=kubernetes.io/cluster/gco-<region> \
  --query 'Volumes[].{V:VolumeId,Size:Size,State:State}'
```

### Observed

Two independent teardowns produced the identical result — same two volumes, same
sizes, both `available`:

```
50 GiB  vol-…  tag kubernetes.io/created-for/pvc/name = prometheus-…-db
 5 GiB  vol-…  tag kubernetes.io/created-for/pvc/name = alertmanager-…-db
```

Both carried `kubernetes.io/cluster/gco-<region>: owned` and
`KubernetesCluster: gco-<region>`. 55 GiB of gp3 is roughly **$4.40/month**, per
destroyed region, indefinitely.

Two for two suggests this is the normal outcome of a deploy/destroy cycle rather
than an edge case. Note the volumes carry the CSI driver's
`kubernetes.io/cluster/*` tags and **not** the CDK `Project=GCO` tag, so they are
invisible to any cleanup or cost query scoped by project tag.

### Expected

`gco stacks destroy` removes the storage it caused to be created, or names it
explicitly so the operator can decide. Silent retention is the problem: the
command reports success and there is no indication anything survived.

### Why this fits the existing destroy path

The destroy already performs exactly this class of AWS-side cleanup, through the
`record_cleanup(...)` registry in `cli/stacks.py` — orphaned bastions (terminated
instances), bastion IAM, the backup vault, the EKS security group. Dynamically
provisioned PVs look like the same shape of problem: a resource the stack caused
to exist that CloudFormation does not own and therefore does not delete.

### Suggested fix

Add a cleanup step, registered alongside the existing ones, that after the
cluster is gone enumerates volumes tagged `kubernetes.io/cluster/<cluster>` and
either deletes them or reports them:

```python
record_cleanup("dynamic-pvs", self._cleanup_cluster_volumes(cluster))
```

Two details worth deciding explicitly:

1. **Delete vs report.** Deleting a volume is irreversible, and someone may want
   Prometheus history. A `--retain-volumes` / `--delete-volumes` flag (or
   defaulting to report-and-list, with deletion opt-in) keeps the choice with the
   operator. Either way, *saying* what survived is the minimum.
2. **Ordering.** The sweep has to run after the cluster is deleted, otherwise the
   CSI driver may recreate or re-attach while it is still reconciling.

A smaller alternative, if cleanup is out of scope: have destroy print any
surviving cluster-tagged volumes at the end, so the leak is visible at the moment
it is created rather than discovered on a bill.

### Workaround

```bash
aws ec2 describe-volumes --region <region> \
  --filters Name=tag-key,Values=kubernetes.io/cluster/gco-<region> \
            Name=status,Values=available \
  --query 'Volumes[].VolumeId' --output text
# then, after confirming the cluster is gone and each volume is detached:
aws ec2 delete-volume --region <region> --volume-id <vol-id>
```

Verify the cluster really is absent first (`aws eks describe-cluster --name
gco-<region>` → `No cluster found`), since a detached volume belonging to a live
cluster may simply be between pod restarts.

sagarneeldubey:
Add command-aware EBS volume disposition to regional stack destruction with fail-closed identity and cluster-absence gates, exact tag-scoped discovery, just-in-time deletion safety rechecks, structured per-volume outcomes, and shared reporting for single and orchestrated destroy paths.

- CLI policy resolver with `--retain-volumes`/`--delete-volumes` and `destroy-all -y` implicit delete authorization (`cli/volume_cleanup.py`, `stacks_cmd.py`)
- Region/cluster-tag scoped discovery, absence verifier, and deletion service
- Shared cleanup renderer and exit-status mapping
- StackManager single and orchestrated regional cleanup integration
- Strict-validation PVC/PV/EBS inventory, isolated retain-override and implicit-delete lifecycles, and fixture cleanup with final EBS accounting
- 13 Hypothesis correctness properties plus unit and mocked integration tests
- Updated CLI help, docs, and validation documentation
- **Follow-up fix:** single-stack `destroy --delete-volumes` now resolves its regional target from the stack name itself instead of live `deployment_regions` config, so cleanup still runs when an external caller has already detached the region before destroying it (see "Follow-up fix" below)

## Summary

Deleting an EKS cluster does not delete the PersistentVolumes its EBS CSI driver provisioned, so every deploy/destroy cycle leaves `available` EBS volumes behind — tagged `kubernetes.io/cluster/<cluster>: owned` for a cluster that no longer exists, unable to ever reattach, and billing indefinitely. With the shipped defaults this is the observability stack (kube-prometheus-stack provisions ~50 GiB Prometheus and ~5 GiB Alertmanager PVCs). Because these volumes carry the CSI driver's `kubernetes.io/cluster/*` tags rather than the CDK `Project=GCO` tag, they are invisible to project-scoped cleanup and cost queries.

This PR adds an EBS volume cleanup stage to regional stack destruction, registered alongside the existing `record_cleanup(...)` steps in `cli/stacks.py` (bastion, backup vault, EKS security group). It is fail-closed: it acts only on an exact configured `<project>-<region>` regional stack, only after definitive stack deletion and verified EKS cluster absence, and only on volumes carrying the exact cluster tag key in the exact target Region. Ownership, `available` state, and zero-attachment are revalidated just-in-time immediately before each `DeleteVolume`. Stack-deletion success and volume-cleanup success are tracked separately so an already-deleted stack is never relabeled as failed when storage cleanup is incomplete.

Disposition is operator-controlled: single-stack destroy and interactive `destroy-all` default to **retain** (report-and-warn), `--retain-volumes` is an explicit non-destructive override, and deletion is opt-in via `--delete-volumes` (with a dedicated irreversible-data prompt when non-interactive confirmation is absent) or implicitly authorized by `destroy-all -y` unless `--retain-volumes` is supplied. Every path emits a complete structured per-volume outcome and summary through the existing cleanup reporting channel.

Per the maintainer request on the issue, the behavior is also wired into the strict live-release-validation workflow as isolated retain-override and delete lifecycles so it is preserved going forward.

### Follow-up fix: single-stack destroy silently skipped cleanup after an external `remove_deployment_region`

Live integration testing (an external tool that drives `destroy_stack` over the MCP interface) found that single-stack `gco stacks destroy --delete-volumes` could silently do nothing: its target resolution required the stack's Region to still be listed in `cdk.json`'s live `deployment_regions.regional`, but that same caller's recommended per-region teardown ordering calls `remove_deployment_region` *before* `destroy_stack` — specifically to release the control plane's Global Accelerator ENIs first. By the time `destroy_stack --delete-volumes` ran, the Region was already gone from config, so cleanup resolved to "not regional" and exited with no discovery, no deletion, and no log output at all. The Prometheus/Alertmanager PVs this feature exists to catch survived a "successful" destroy.

The fix adds `resolve_regional_volume_target_from_stack_name()` (`cli/volume_cleanup.py`), used only by the single-stack destroy path: it resolves the regional target from the stack name the operator explicitly asked to destroy (still validated against known CloudFormation Regions — never an arbitrary string), rather than consulting live config. The command itself, along with `--delete-volumes`, is already the operator's authorization; requiring the same Region to *also* still be configured a moment later penalizes exactly the ordering this tool recommends. Orchestrated `destroy-all` is unaffected — it already captures its config-based resolution once, before any of its own stacks are destroyed, so it never hits this timing gap. Also threads a `delete_volumes` parameter through the `destroy_stack` MCP tool so callers driving GCO over MCP (not the bare CLI) can request the same cleanup.

## Type of change

Suggested label: `enhancement` (matches the linked issue).

## Testing

- [x] `pytest tests/` passes locally
- [x] `cdk synth` succeeds (if CDK code changed)
- [ ] If `examples/` changed: `gco examples validate --static-only` passes, and a live `gco examples validate --examples <changed>` run backs any behavior change

**Full local suite** (`pytest tests/ -n auto`, macOS, Python 3.14.6, `pip install -e ".[dev,mcp]"` incl. `aws-cdk-lib==2.263.0`, Lambda assets prebuilt via `prepare_cdk_assets`):

```
10963 passed, 122 skipped, 3 failed in 1444.81s (24:04)
```

The EBS cleanup feature and the entire remainder of the suite are green. The 3 failures are pre-existing and environmental — `git diff` against the merge-base confirms none of the failing test files or their production modules were modified on this branch:

- `test_storage_mcp.py::...::test_regional_upload_is_gated_and_uses_private_descriptor_snapshot` — **Linux-only.** `gco_mcp/local_data.py` stages uploads via `/dev/fd/<dirfd>/<name>` descriptor traversal, which resolves on Linux (`/dev/fd` → `/proc/self/fd`) but not on macOS.
- `test_mcp_destructive_gating.py::TestModelUploadGating::test_models_upload_argv_minimal` — same Linux-only `/dev/fd` mechanism.
- `test_analytics_bucket_isolation_property.py::...::test_regional_job_pod_role_isolates_cluster_shared_from_studio_only` — **not a correctness failure;** a Hypothesis "unreliable test timings" flake where the first cold CDK synth on macOS took 36.7s (>20s deadline) but 3.96ms on rerun. Stays under the deadline on CI's Linux runners.

**Follow-up fix commit** — the affected suites re-run clean after the single-stack resolution fix (`pytest tests/ -k "volume_cleanup or mcp_tasks"`):

```
451 passed in ~20s
```

**Live AWS verification for the follow-up fix:** provisioned a real regional stack (`gco-us-west-2`), confirmed all 4 dynamically-provisioned volumes (Prometheus 50GiB, Alertmanager 5GiB, Grafana 10GiB, MLflow 10GiB) `in-use`, then destroyed via the exact `remove_deployment_region` → `destroy_stack --delete-volumes` ordering that previously no-op'd. The destroy task log now shows the full discovery/eligibility/deletion report for all 4 volumes (it previously logged nothing), and post-teardown `aws ec2 describe-volumes` plus an independent residual-resource scan both confirm zero volumes remain in the Region. This is an ad hoc manual verification distinct from the formal live-release-validation workflow described below, which remains pending.

Notes:
- `cdk synth` path is exercised locally via the CDK synthesis/config-matrix tests (they pass) after `aws-cdk-lib` install and Lambda-asset prep. This PR changes no CDK/`gco/stacks/` code.
- All 13 correctness properties run as dedicated Hypothesis tests (≥100 examples each) with the `Feature: cleanup-dynamic-ebs-volumes, Property N: ...` tags.
- No ordinary test touches live AWS — all AWS interactions are mocked/injected, except the ad hoc manual verification described above.
- Repository quality gates run clean locally: `ruff check`, `ruff format --check`, strict `mypy`, `mkdocs build --strict`, markdownlint, and spec diagnostics.
- A truly 100% green full suite is a Linux-only outcome (the two `/dev/fd` tests are platform-bound), which is why CI runs on `ubuntu-latest`.

## Live release validation

- [ ] Not required (explain the applicability decision below)
- [ ] Required and completed locally with `--actions all` for this exact SHA; a sanitized `PASSED` summary comment (run ID, SHA, per-action statuses) was posted on the pull request
- [x] Required but pending explicit validation-account and KMS-deletion authorization

**Applicability rationale:** This change introduces a new destructive, live-AWS runtime operation (`DeleteVolume` on cluster-tagged EBS volumes) on the deployed teardown path, which falls in the risk-based "live-resource CLI change" category that still warrants validation. The strict-validation lifecycles (`retain-override` and `delete`) are implemented and covered by mocked scenario/evidence/final-inventory tests, but the live destructive run via the formal `gco release validate` workflow was intentionally not executed here: it is identity-fenced and requires an explicitly selected validation account and explicit EBS-fixture-cleanup authorization. It should be run against this exact SHA (`3608c23`) before merge. Separately, the follow-up fix commit was manually verified end-to-end against a real deployed stack — see Testing above — but that ad hoc run does not substitute for the gated `--actions all` / `--volume-scenario both` validation requested here.

**Sanitized validation summary or comment link (never post the full reports or `checkpoint.json`; they carry account IDs, ARNs, and endpoint URLs):** Pending — a sanitized `PASSED` summary (run ID, SHA, per-action statuses) will be posted after the identity-fenced `--volume-scenario both` run completes in the validation account.

## Checklist

- [x] Documentation updated (README, `docs/CLI.md`, `docs/LIVE_RELEASE_VALIDATION.md`, validation README, inline docstrings)
- [x] No secrets, credentials, or customer data committed (a synthetic test placeholder key flagged by the secret scanner was reworked so it no longer matches a credential pattern)
- [ ] `requirements-lock.txt` regenerated if `pyproject.toml` changed — **N/A**, `pyproject.toml` was not modified
- [x] Changes align with the architecture described in `docs/ARCHITECTURE.md` (reuses the existing `record_cleanup(...)` registry and post-cluster-deletion ordering)

## Related issues

Fixes #268


Jmevorach:
Thanks for this, @sagarneeldubey; command-aware volume disposition with fail-closed gates is exactly the right shape for the destroy path, and the Hypothesis properties are a great touch.

One sequencing note: #277 is in final validation and lands first; it reworks a lot of the same ground (`cli/stacks.py` destroy/cleanup paths, the live-validation harness). We'll start the full review here as soon as that merges, and this branch will need a rebase onto main at that point rather than a merge-up, since the overlap is real. Appreciate the patience; this is queued, not parked.

Jmevorach:
Hi @sagarneeldubey — good news: #277 merged this morning (`fb31f87`), so this is next up. I'm planning to start the detailed review next week; if the branch is rebased by then I can dive straight into the substance.

GitHub is showing conflicts against `main` now, which we expected — the two PRs share the destroy path. Could you rebase onto latest `main` when you get a chance? The hotspots you'll hit, to save you some archaeology:

- **`cli/stacks.py`** — `destroy_orchestrated`'s `finish()` funnel now also runs a success-only traffic-dial SSM parameter purge (`_cleanup_traffic_dial_parameters`). It follows the `record_cleanup(...)` outcome-dict pattern, so your volume-cleanup hook should slot in right next to it.
- **`tests/conftest.py`** — the autouse destroy-cleanup no-op list gained that new helper (see `_DESTROY_CLEANUP_OWNERS`). If your volume cleanup adds its own boto3-backed helper, wiring it into the same list keeps the orchestration tests hermetic — your barrier tests will likely want an owner-class entry there too.
- **`scripts/live_release_validation/actions/{destroy,final_inventory}.py`** — final-inventory gained an expired-DynamoDB-stream acceptance, so your volume-inventory extension will want to sit on top of that shape.

Happy to pair on the rebase and/or any edits I end up recommending — just say the word. And thanks again for the thorough property coverage in here; it's going to make the review a pleasure.

sagarneeldubey:
@Jmevorach No worries I got it. I am also testing a real world scenario by allowing gco destroy to tear down actual EBS volumes created by itself.

sagarneeldubey:
Thanks @Jmevorach — the pointers saved real archaeology. Rebase is done and the three hotspots are addressed in `a65c712`.

One correction on the conflict state: the branch was already rebased onto latest `main` shortly after your comment, so GitHub is reporting `MERGEABLE` now and the merge base is `061b5a5` (v6.4.0, which includes #277). `BLOCKED` is just the missing approval. The mechanical rebase was clean, so nothing conflicted — but that also meant it silently did none of the integration below, which is where the actual work was.

**`cli/stacks.py` — the traffic-dial seam.** The volume barrier was already publishing through `record_cleanup` and already feeding `finish(overall)`, so a failed cleanup withholds your success-only purge exactly as a surviving stack does. Nothing proved it though, so there are now two tests pinning both directions: an EC2 delete failure with every stack deleting cleanly holds the purge back (and still reports no failed stack), and the fully-clean teardown calls it once.

On placement: I did not move it into `finish()`. It publishes through the same funnel, but it is a gate *within* teardown rather than an exit sweep — cluster identity is only readable at that point, and an unaccounted volume has to stop the global phase, which a `finish()` sweep runs too late to do. Moving it would also break the ordering contract (`watchdogs < publications < global destroy`). I left a comment at the call site so the next reader does not have to re-derive that. Happy to revisit if you see it differently.

**`tests/conftest.py` — this one was a genuine hole, thanks for flagging it.** The feature reaches AWS through exactly one seam, `_volume_cleanup_clients`, whose fallback builds real boto3 EKS/EC2 clients. It was unpatched, so a future test passing `volume_cleanup_request` without injecting dependencies would have issued live `DescribeVolumes`/`DeleteVolume` against whatever `project_name` resolved to. There is now an autouse guard that keeps the injected-factory path untouched and removes only the fallback, so that test fails loudly instead.

Two deliberate departures from `_DESTROY_CLEANUP_OWNERS`:

- It is not a blanket no-op. The barrier tests exercise the real barrier through injected fakes, so no-oping it would silently neuter them — the opposite of what the list is for.
- Opt-out is a marker, not a class name. Every volume test in the suite is a module-level function, so `request.cls` is `None` and a class entry can never match. The marker works for both styles; I kept the class set alongside it so a future `Test...` class can opt out the way the destroy-cleanup owners do. The one test that genuinely owns the fallback (`test_default_dependencies_create_one_region_scoped_boto3_client`) now carries it — and it was the guard catching that test that confirmed the guard works.

**`final_inventory.py` — sitting on the acceptance shape.** Your stream acceptance set the pattern: a tolerated leftover is disclosed as a flat `accepted_*` list whose records name the API observation that established them. Volumes already tolerated in-flight deletions (`disposition == "deleting"` is not residual and does not fail the run) but only reported it nested inside the per-Region accounting — the silent tolerance the pattern exists to prevent. `final-inventory` now also returns `accepted_pending_deletion_volumes`, reprojected from the observation it already made rather than re-describing, so the disclosure cannot disagree with the accounting it came from, and uniformly `[]` when nothing was tolerated. `docs/LIVE_RELEASE_VALIDATION.md` documents it next to the stream class.

One thing I noticed while in there and deliberately did **not** fix, since it is pre-existing and outside this PR's scope: `destroy.py`'s already-destroyed and resume-after-absence early returns never capture volume targets, while `action_final_inventory` calls `volume_residual_inventory` unconditionally. A resumed run can therefore raise on fixtures no resume path was given a chance to clean. `_checkpoint_retained_kms_keys` is called on all three paths, so KMS evidence survives a resume and volume evidence does not. Happy to fix it here if you'd rather, or file it separately.

Also worth knowing before you review: `tests/test_mcp_destructive_gating.py::TestModelUploadGating::test_models_upload_argv_minimal` and `tests/test_storage_mcp.py::TestMCPStorageTools::test_regional_upload_is_gated_and_uses_private_descriptor_snapshot` fail with `KeyError: 'args'` on clean `main` too — I verified in a worktree at `061b5a5`, so they are not from this branch. Everything else is green (11308 passed; three further failures only under parallel load, all passing in isolation).

Taking you up on the offer to pair — happy to walk the barrier ordering and the `_regional_volume_cleanup_outcome` fail-closed paths whenever suits you.

sagarneeldubey:
Correction to my last comment, so you don't go chasing a phantom: those two MCP tests (`test_models_upload_argv_minimal`, `test_regional_upload_is_gated_and_uses_private_descriptor_snapshot`) **pass in CI**. `unit:pytest:core` is green on both shards for `a65c712`. They reproduce only on my machine, and they do so on clean `main` as well, so it's local environment rather than anything in the repo or this branch. Nothing for you to look at there.

All seven workflows are green on `a65c712`: Unit Tests, Linting (including `mypy:strict` and `mypy:stacks`), Integration Tests, Security, Floci, Mooncake, Inference Streaming Proxy. Ready for review whenever you get to it.

Jmevorach:
Thanks for taking this on, @sagarneeldubey. The leak is real, and the mechanics you landed on are right; one default I'd flip, but otherwise what follows is about size and shape, not correctness.

I'm going to close this and pick it up on a separate branch, with substantially less code.

## What carries over

These are the hard part of the problem and I'm keeping them:

- **Sweep only after the cluster is confirmed gone**, so the CSI driver isn't still reconciling.
- **Exact tag scoping** on `kubernetes.io/cluster/<cluster>` in the target Region; not the `Project=GCO` tag, which these volumes don't carry.
- **Just-in-time recheck before each delete** rather than trusting the discovery snapshot.
- **Stack-deletion success tracked separately from cleanup success**, so an already-deleted stack never gets relabeled as failed.

Your follow-up fix is also a good catch: `remove_deployment_region` before `destroy_stack` is the ordering we recommend, and cleanup silently no-op'ing under it would have made the feature useless in exactly the path people follow.

## One behavior I'm changing: delete by default

Retain-by-default is the defensible reading of "`DeleteVolume` is irreversible," but it puts this storage at odds with every other stateful resource we own. `gco/stacks/` uses `RemovalPolicy.DESTROY` 65 times against 3 `RETAIN` (a configurable analytics opt-in, a live-validation log group behind a context flag, and the EKS secrets KMS key). All 10 `auto_delete_objects` are `True`, so buckets go with their contents. `_cleanup_backup_vault` already deletes backup recovery points on destroy; real backup data, no dedicated flag, no second prompt.

More to the point, `gco-observability-gp3` declares `reclaimPolicy: Delete`. That is already the stated intent for these exact volumes. They only survive because the cluster is deleted before the PVCs are, so the CSI driver never receives the delete event. Deleting them honors the policy the StorageClass already declares; retaining them silently contradicts it, and leaves the operator paying for storage in a Region they explicitly tore down.

So: delete by default, with `--retain-volumes` as the opt-out. The existing `destroy` confirmation is the authorization; a second volume-specific prompt would be inconsistent with a command that already drops bucket contents and backup recovery points without one. That also removes the need for `DeletionAuthorizationSource` and the implicit-`-y` rule.

## Why close instead of iterating

The diff is +24,082 / −23 across 76 files. The six cleanup steps already in the destroy path; orphaned bastions, bastion IAM, backup vault, EKS security groups, implicit log groups, traffic-dial parameters; are 28 to 152 lines each, 494 lines for all six combined. This adds ~3,800 production lines for a seventh, including `cli/volume_cleanup.py` at 2,252 lines and 34 classes, plus 16,808 test lines across 48 files.

The clearest symptom: `_configured_regional_volume_regions()` enforces a rule that cleanup must never infer a Region from a stack name, with a fail-closed `regional-configuration-unreadable` path behind it; and then the follow-up fix had to add `resolve_regional_volume_target_from_stack_name()` to do exactly that inference. Meanwhile `_cleanup_eks_security_groups`, the shipped cleanup step next door in the same file, already derives both values in two lines:

```python
region = region or stack_name.replace(f"{project_name}-", "", 1)
cluster_name = stack_name
```

That's the whole mechanism, on the ordinary path, for a step that also deletes real AWS resources. Deriving identity from the stack name is convention here, not a hazard. Working 24k lines into that shape through review rounds is slower than writing the small version, and the rounds would go to scaffolding rather than the behavior you already got right.

## On reviewability

Worth stating plainly since it shapes what works here: I read every line of every PR I merge. Nothing gets waved through on a green CI run and a summary. So the size of a diff is a claim on review budget, and 24k lines is more than this feature can justify spending.

For future PRs, the thing that helps most is commits a reviewer can actually follow; each one scoped to a single change, with a message explaining why, in an order that tells the story of the work. If a change genuinely needs to be large, splitting it across PRs that each stand on their own beats one that has to be read all at once.

## What I'm building

One `_cleanup_cluster_volumes` method in `cli/stacks.py` registered as `record_cleanup("dynamic-pvs", ...)`, returning the same details dict the neighboring steps return. Tag-and-`available`-filtered `describe_volumes`, skip anything attached, recheck before each delete, treat `InvalidVolume.NotFound` as success, one volume's failure doesn't stop the rest. A single `--retain-volumes` flag on `destroy` and `destroy-all` to opt out. Either way the details dict names every volume with its ID, size, and AZ, so what was deleted or left behind is on the record; silence is the actual bug in #268. Reporting through the existing cleanup channel, tests next to the existing cleanup tests, plus the e2e coverage I asked for on the issue. Target is ~150 lines and no new module.

Keeping a `retain_volumes` parameter on the `destroy_stack` MCP tool as well. I'll link the branch here; genuinely useful find, and the live verification you did on the teardown ordering saved me a bug.

Jmevorach:
Thanks for taking this on, @sagarneeldubey. The leak is real, and the mechanics you landed on are right — one default I flipped, but otherwise what follows is about size and shape, not correctness.

Closing this in favour of #297, which shipped in **v6.5.0** and closes #268. It carries the behaviors listed below, so the substance of your work is in the release; what changed is the amount of code around it.

## What carries over

These are the hard part of the problem and they are all in the shipped version:

- **Sweep only after the cluster is confirmed gone**, so the CSI driver isn't still reconciling.
- **Exact tag scoping** on `kubernetes.io/cluster/<cluster>` in the target Region — not the `Project=GCO` tag, which these volumes don't carry.
- **Just-in-time recheck before each delete** rather than trusting the discovery snapshot.
- **Stack-deletion success tracked separately from cleanup success**, so an already-deleted stack never gets relabeled as failed.

Your follow-up fix was also a good catch: `remove_deployment_region` before `destroy_stack` is the ordering we recommend, and cleanup silently no-op'ing under it would have made the feature useless in exactly the path people follow.

## One behavior I changed: delete by default

Retain-by-default is the defensible reading of "`DeleteVolume` is irreversible," but it puts this storage at odds with every other stateful resource we own. `gco/stacks/` uses `RemovalPolicy.DESTROY` 65 times against 3 `RETAIN` (a configurable analytics opt-in, a live-validation log group behind a context flag, and the EKS secrets KMS key). All 10 `auto_delete_objects` are `True`, so buckets go with their contents. `_cleanup_backup_vault` already deletes backup recovery points on destroy — real backup data, no dedicated flag, no second prompt.

More to the point, `gco-observability-gp3` declares `reclaimPolicy: Delete`. That is already the stated intent for these exact volumes. They only survive because the cluster is deleted before the PVCs are, so the CSI driver never receives the delete event. Deleting them honors the policy the StorageClass already declares; retaining them silently contradicts it, and leaves the operator paying for storage in a Region they explicitly tore down.

So it ships as delete-by-default, with `--retain-volumes` as the opt-out. The existing `destroy` confirmation is the authorization — a second volume-specific prompt would be inconsistent with a command that already drops bucket contents and backup recovery points without one. That also removes the need for `DeletionAuthorizationSource` and the implicit-`-y` rule.

## Why close instead of iterating

The diff is +24,082 / −23 across 76 files. The six cleanup steps already in the destroy path — orphaned bastions, bastion IAM, backup vault, EKS security groups, implicit log groups, traffic-dial parameters — are 28 to 152 lines each, 494 lines for all six combined. This adds ~3,800 production lines for a seventh, including `cli/volume_cleanup.py` at 2,252 lines and 34 classes, plus 16,808 test lines across 48 files.

The clearest symptom: `_configured_regional_volume_regions()` enforces a rule that cleanup must never infer a Region from a stack name, with a fail-closed `regional-configuration-unreadable` path behind it — and then the follow-up fix had to add `resolve_regional_volume_target_from_stack_name()` to do exactly that inference. Meanwhile `_cleanup_eks_security_groups`, the shipped cleanup step next door in the same file, already derives both values in two lines:

```python
region = region or stack_name.replace(f"{project_name}-", "", 1)
cluster_name = stack_name
```

That's the whole mechanism, on the ordinary path, for a step that also deletes real AWS resources. Deriving identity from the stack name is convention here, not a hazard. Working 24k lines into that shape through review rounds is slower than writing the small version, and the rounds would go to scaffolding rather than the behavior you already got right.

## On reviewability

Worth stating plainly since it shapes what works here: I read every line of every PR I merge. Nothing gets waved through on a green CI run and a summary. So the size of a diff is a claim on review budget, and 24k lines is more than this feature can justify spending.

For future PRs, the thing that helps most is commits a reviewer can actually follow — each one scoped to a single change, with a message explaining why, in an order that tells the story of the work. If a change genuinely needs to be large, splitting it across PRs that each stand on their own beats one that has to be read all at once.

## What shipped

One `_cleanup_cluster_volumes` method in `cli/stacks.py` registered as `record_cleanup("dynamic-pvs", ...)`, returning the same details dict the neighbouring steps return, plus a single `--retain-volumes` opt-out on `destroy` and `destroy-all` and a `retain_volumes` parameter on the `destroy_stack` MCP tool. +1,214 / −3 across 16 files, most of it tests and docs.

Two things worth reporting back, since both came from your issue:

**The live validation reproduced your report exactly.** Four CSI volumes were provisioned. Grafana and MLflow deleted themselves — plain `Deployment` PVCs go with the namespace and `reclaimPolicy: Delete` does the rest. Prometheus (50 GiB) and Alertmanager (5 GiB) did not, because StatefulSet `volumeClaimTemplates` PVCs are retained by design, so the driver never receives a delete event. Both stranded as `available` at exactly the 55 GiB you reported. So your two volumes were not an edge case, and now we know why it is those two and not all four.

**Your e2e request found a second gap.** The validation harness had no EBS volume awareness at all, and its residual scan keys off the CloudFormation stack-name tag, a `gco:project` tag, or a project-prefixed name — none of which these volumes carry. `final-inventory` would have certified a clean teardown with volumes still billing: the same blind spot you documented for cost queries, sitting in the gate that decides whether a run passed. There is now a `cluster_volumes` scanner, so a stranded volume fails the all-zero gate.

Also worth crediting: deriving the target from the stack name rather than live config, which is what your follow-up fix pointed at, is what makes the `remove_deployment_region` → `destroy_stack` ordering safe by construction rather than by a second code path.

## Follow-up

One known item is deferred to a patch: EKS Auto Mode node root volumes carry the same `kubernetes.io/cluster/*` tag, and while the runtime sweep skips them correctly (it requires `available` plus zero attachments, and they are `DeleteOnTermination`), the new `cluster_volumes` scanner applies no state filter, so a node volume transiting `deleting` could fail the all-zero gate spuriously. It did not fire in the validation run and affects only the harness rather than shipped behavior, so it goes out as a patch with its own green-SHA e2e cycle — and I'd like your review on that one.

Thanks again for finding and chasing this. The leak was real, the live verification you did on the teardown ordering saved me a bug, and none of the size feedback should read as anything other than what it is.

PR #297, opened 2026-08-24
feat: delete orphaned CSI-provisioned EBS volumes on stack destroy

### LINKED ISSUE #268: [feat] `gco stacks destroy` leaves the cluster's dynamically-provisioned EBS volumes behind

Jmevorach:
## Summary

Deleting an EKS cluster does not delete the PersistentVolumes its EBS CSI driver provisioned. Every deploy/destroy cycle therefore stranded `available` volumes tagged `kubernetes.io/cluster/<cluster>` for a cluster that no longer exists — nothing can reattach them, and they bill indefinitely. With the shipped defaults that is the observability stack (Prometheus, Grafana, Alertmanager, MLflow). Because the volumes carry the CSI driver's tags rather than the CDK `Project` tag, no project-scoped cleanup or cost query could see them.

This adds a seventh step to the existing `record_cleanup(...)` registry in `cli/stacks.py`, alongside orphaned bastions, bastion IAM, the backup vault, implicit log groups, traffic-dial parameters, and EKS security groups.

**Behavior:** volumes are deleted by default; `--retain-volumes` lists them instead. Either way every volume is named with its ID, size, type and AZ, because silent retention is the actual defect in #268.

Deletion is the default because it honors intent already declared elsewhere. The `gco-observability-gp3` StorageClass sets `reclaimPolicy: Delete`; these volumes survive only because the cluster is torn down before its PVCs are, so the driver never receives the delete event. It also matches how the rest of the project treats stateful resources — `gco/stacks/` uses `RemovalPolicy.DESTROY` 65 times against 3 `RETAIN`, all 10 `auto_delete_objects` are `True`, and `_cleanup_backup_vault` already deletes backup recovery points on destroy without a dedicated flag.

### Where it runs

- **`destroy-all`** — at the post-regional barrier, where `_destroy_phase_remaining_stacks` has already confirmed every regional stack is absent. Running there rather than inside the per-stack loop means parallel and sequential teardowns publish the same outcomes in the same order, with no concurrent access to the report.
- **`destroy <stack>`** — this path calls `StackManager.destroy` directly and has no cleanup registry, so it previously had no sweep at all. That is the exact command #268 reproduces with, so it now sweeps after a reported success.

Cleanup results deliberately do not feed the `failed` list. Those stacks are already deleted, and relabeling one would send the CLI's 3-attempt retry loop back to CDK for a stack that no longer exists.

### Safety

Modelled on the neighbouring `_cleanup_eks_security_groups`:

- Proves the cluster is absent (`describe_cluster` → `ResourceNotFoundException`) before issuing any EC2 call, so a still-reconciling CSI driver is never raced.
- Scopes discovery to the exact `kubernetes.io/cluster/<cluster>` tag key and `status=available` in the exact target Region.
- Rechecks ownership tag, `available` state and zero attachments immediately before each `DeleteVolume` rather than trusting the discovery snapshot. An unreadable volume is reported as blocked, not deleted.
- Treats `InvalidVolume.NotFound` as success, lets one volume's failure not stop the others, and never raises into the destroy flow.

### Pricing

Retained volumes are priced against the target Region at teardown time through the Price List API, per volume type, rather than from a constant in the source. EBS rates differ by Region and change over time, so a checked-in number would quietly drift into misinforming the operator. When the rate cannot be established — no `pricing:GetProducts`, an unroutable endpoint, an emulator without the API — the command says the cost is unknown instead of printing a stale figure. Timeouts are short and retries few; this is a cosmetic annotation on the teardown path and must never hold it up.

### Credit

The design carries over the behaviors @sagarneeldubey landed on in #281 — post-cluster-absence ordering, exact tag scoping, the just-in-time recheck, and keeping cleanup failures from relabeling a deleted stack. #281 also found, through live testing, that the recommended `remove_deployment_region` → `destroy_stack` ordering could make cleanup silently no-op; deriving the target from the stack name (as `_cleanup_eks_security_groups` already does) avoids that class of bug entirely.

## Type of change

- [x] `breaking:` Breaking change (major version bump)

On breaking-ness: the prior behavior was the defect, not a contract — `reclaimPolicy: Delete` already declared the intent. Anyone relying on the old outcome gets `--retain-volumes`. Worth a release-note callout so nobody is surprised.

## Testing

- [ ] `pytest tests/` full suite: left to CI on this PR
- [x] `ruff check`, `ruff format --check`, strict `mypy` clean on every changed file

**Unit + orchestration** (`tests/test_stacks_extended_coverage.py`, `tests/test_stacks_destroy_hardening.py`, `tests/test_cli_main.py`, `tests/test_mcp_tasks.py`): `TestClusterVolumeCleanup` sits beside `TestEksSecurityGroupCleanup` and matches its shape — absence gate blocking every EC2 call, discovery scoping, deletion, retain reporting, the recheck sparing a reattached volume and one whose ownership tag vanished, `InvalidVolume.NotFound` as success, one failure not stopping the rest, swallowed top-level errors. Pricing is covered both ways: queried for the teardown Region and volume type, and reported unavailable rather than guessed when it fails.

**Floci layer** (`tests/test_floci_volume_cleanup.py`, 6 tests, verified green against a local `floci/floci:1.7.0`): the real method against real emulator state, which proves what `MagicMock` cannot — that the `tag-key` + `status` filter pair actually selects the right volumes server-side, that `DeleteVolume` really removes them (verified by re-reading EC2, not by asserting on a mock), that another cluster's volumes and untagged volumes are untouched, and that an absent cluster really does raise `ResourceNotFoundException` while a live one stops the sweep before any deletion. Floci has no Price List API, so its retain test is genuine coverage of the pricing fallback.

**Live-release-validation gate** (`scripts/live_release_validation/inventory/`): per the request on #268 to make sure this cannot be undone silently later. The harness had **no EBS volume awareness at all**, and its residual scan matches on the CloudFormation stack-name tag, a `gco:project` tag, or a project-prefixed name — none of which CSI-provisioned volumes carry. So `final-inventory` would have reported zero residual project resources while billable volumes survived: the same blind spot #268 describes for project-scoped cost queries, in the gate that authorizes calling a teardown clean.

A `cluster_volumes` scanner now resolves ownership from the cluster name inside the tag key, so another project's cluster in the same account is never claimed. Registering the category and scanner is all that was needed — `final-inventory` already fails on any non-zero count. Volume state is deliberately not filtered, since a detaching or errored volume is just as residual as an available one. The Floci harness-inventory test was extended so a real cluster-tagged volume must be seen while it exists, must keep the all-zero gate *failing* after every stack- and project-tagged resource is deleted, and only then may absence be proven.

**Safety fixture:** both new classes join `_DESTROY_CLEANUP_OWNERS`, and the autouse `_no_real_destroy_cleanup_aws_calls` fixture now stubs `_cleanup_cluster_volumes` for every other test. Without it, any orchestration test naming a live regional stack would issue real `DeleteVolume` calls — the most destructive helper that fixture guards. This was caught by the Floci tests failing against the stub, which is the fixture working as designed.

`requirements-lock.txt` not regenerated: no dependency changes.

## Live release validation

- [x] Required and completed for this exact SHA (`c3310cb`); sanitized `PASSED` summary posted on this PR

**Applicability rationale:** this introduces a new destructive live-AWS operation (`DeleteVolume`) on the teardown path, which is the "live-resource CLI change" category.

**Result:** all 12 actions passed against a real account (2h51m). The deploy provisioned four CSI volumes; Grafana and MLflow deleted themselves with their namespace, while Prometheus (50 GiB) and Alertmanager (5 GiB) stranded as `available` — the exact two volumes and 55 GiB total from #268, reproduced independently. The sweep deleted both once cluster absence was proven, and a post-teardown scan found zero cluster-tagged volumes. `final-inventory` passed all-zero including `cluster_volumes: 0` from the new scanner. Node root volumes, which share the same tag, were correctly skipped as attached. Sanitized summary is in a comment on this PR.

**New IAM requirement:** the sweep calls `ec2:DeleteVolume` and, for retained volumes only, `pricing:GetProducts`. Operator credentials running `gco stacks destroy` need both; the pricing call degrades to "cost unknown" without permission rather than failing.

## Checklist

- [x] Documentation updated (`docs/CLI.md`, `tests/README.md`, inline docstrings)

Closes #268


Jmevorach:

Sanitized summary. Full reports and `checkpoint.json` stay local — they enumerate account identifiers, ARNs, and endpoint URLs.

- **Run ID:** `20260824T1330Z-c3310cb789d9`
- **SHA:** `c3310cb789d926ddb6bc353a1236575bdc2a8656`
- **Branch:** `feat/cleanup-orphaned-cluster-volumes`
- **Profile:** `configured` (regional `us-east-1`; global / api-gateway / monitoring `us-east-2`)
- **Actions:** `--actions all` (12 selected, 12 passed, 0 failed)
- **Wall clock:** 2h 51m (17:33:23Z → 20:24:45Z)
- **Status:** **PASSED**, `fatal_error: none`

|---|---|---:|
| `preflight` | passed | 36.3s |
| `baseline` | passed | 570.8s |
| `deploy` | passed | 2888.8s |
| `topology` | passed | 536.1s |
| `api` | passed | 24.7s |
| `sqs` | passed | 54.7s |
| `central-queue` | passed | 45.7s |
| `schedulers` | passed | 47.9s |
| `opencost` | passed | 1.2s |
| `convergence` | passed | 21.3s |
| `destroy` | passed | 4769.2s |
| `final-inventory` | passed | — |

(`final-inventory` is recorded as `0.0s` in the report; its substantive pass logged ~569s to the console. No stale-checkpoint reconciliation was recorded.)

### What this proves about the feature

`cluster_observability` was enabled, so the deploy provisioned four real CSI volumes. Their teardown behavior split in a way worth recording:

- **Grafana (10 GiB) and MLflow (10 GiB) deleted themselves.** Plain `Deployment` PVCs go with the namespace, and `reclaimPolicy: Delete` on `gco-observability-gp3` does the rest.
- **Prometheus (50 GiB) and Alertmanager (5 GiB) did not.** Those are StatefulSet `volumeClaimTemplates`, whose PVCs are retained by design, so the CSI driver never receives a delete event. Both landed in `available` — **exactly the two volumes and the same 55 GiB total reported in #268**, reproduced independently.

The sweep then did its job, after the regional stack reached `DELETE_COMPLETE` and cluster absence was proven:

```
gco-us-east-1 is absent from CloudFormation.
Cleaned up 2 orphaned EBS volume(s) (55 GiB) left by cluster gco-us-east-1:
  - vol-… (50 GiB gp3, us-east-1d, pvc=prometheus-…-db)
  - vol-… (5 GiB gp3, us-east-1a, pvc=alertmanager-…-db)
```

A post-teardown `describe-volumes` filtered on `kubernetes.io/cluster/gco-us-east-1` returned **zero** volumes.

Ordering held under real conditions: while the cluster still existed the two volumes sat `available` and were left alone, and the sweep only acted once absence was confirmed. Nothing was over-deleted — the EKS Auto Mode node root volumes share the same `kubernetes.io/cluster/*` tag but were `in-use` with attachments, so the `available` + zero-attachment requirement correctly skipped them, and they went with their instances.

### What this proves about the new validation gate

`final-inventory` passed with `all_absent: true`, `residual: []`, `baseline_differences: []`, and an all-zero residual summary — including **`cluster_volumes: 0`** from the scanner added in this PR. Protected stacks and the ECR baseline were preserved exactly.

### Known follow-up (harness only, no runtime impact)

EKS Auto Mode node root volumes carry `kubernetes.io/cluster/<cluster>` too. The runtime sweep handles them correctly (`available` + zero attachments, and they are `DeleteOnTermination`). The `cluster_volumes` scanner, however, applies no state filter by design, so a node volume transiting `deleting` during teardown could fail the all-zero gate spuriously. It did not fire in this run. Since it affects only the validation harness and not shipped behavior, it is deferred to a patch, which will get its own green-SHA e2e cycle.

PR #299, opened 2026-08-24
fix: ignore volumes EC2 is already deleting in the residual-volume gate

Jmevorach:
## Summary

Follow-up to #297 (shipped in v6.5.0), fixing a false-positive risk in the validation gate that PR introduced.

The `cluster_volumes` scanner counted any volume tagged `kubernetes.io/cluster/<cluster>` as residual regardless of state. That tag is not exclusive to CSI-provisioned PersistentVolumes — **EKS Auto Mode writes it onto node root volumes too**, which live and die with their instances and pass through `deleting` on every teardown. A run that happened to observe one mid-transition would have failed the all-zero `final-inventory` gate for a volume that was already on its way out.

Volumes in `deleting` or `deleted` are now skipped. Every other state still counts — `available`, `in-use`, `creating`, `error` — because after teardown nothing should be holding a cluster-tagged volume at all.

### Scope

**Harness-only. No runtime behavior changes.** The sweep in `cli/stacks.py` was never affected: it requires `available` plus zero attachments, so attached node volumes were always skipped. The live validation of #297 confirmed this empirically — node root volumes (80 GiB and 4 GiB, `in-use`) were left untouched while both stranded PersistentVolumes were deleted.

### How this surfaced

This came out of the live validation run for #297, not from review. Node root volumes sharing the CSI driver's cluster tag is not something the mocked tests could reveal, because the fixtures only ever contained volumes the tests themselves created. It is the kind of thing only a real deploy shows you.

## Type of change

## Testing

- [x] `ruff check`, `ruff format --check`, strict `mypy` clean
- [x] 379 targeted tests pass (`test_live_release_validation`, `..._structure`, `test_stacks_extended_coverage`, `test_stacks_destroy_hardening`, `test_docs_coverage`)

- **Unit:** a test for both terminal states (`deleting`, `deleted`) yielding an empty result, and a parametrized test pinning the four states that still count.
- **Floci:** `TestClusterVolumeResidualScoping` creates a real cluster-tagged volume, asserts the all-zero gate *fails* while it exists, deletes it, waits on `volume_deleted`, and asserts the gate then passes. Verified green against a local `floci/floci:1.7.0` (10/10 floci tests).

`requirements-lock.txt` not regenerated: no dependency changes.

## Live release validation

- [x] Required and completed for this exact SHA (`df94acf`); sanitized `PASSED` summary posted on this PR

**Result:** all 12 actions passed against a real account (2h59m). Mid-teardown the Region held both stranded PersistentVolumes (`available`) and four EKS Auto Mode node root volumes (`in-use`) under the same cluster tag — the exact shared-tag condition this patch addresses. The sweep deleted only the two stranded PVs (55 GiB, same split as the v6.5.0 run but different volume IDs and AZ, so the leak reproduces as a property of StatefulSet `volumeClaimTemplates` rather than one deployment). `final-inventory` passed all-zero with `cluster_volumes: 0`, and a post-teardown scan returns 0.

## Checklist

- [x] Documentation updated (inline docstrings)

Refs #268

### DIFF
--- a/scripts/live_release_validation/inventory/scanners.py
+++ b/scripts/live_release_validation/inventory/scanners.py
@@ -275,6 +275,11 @@ def _ec2_items(client: Any, operation: str, response_key: str) -> list[dict[str,
 
 _CLUSTER_TAG_PREFIX = "kubernetes.io/cluster/"
 
+#: Volume states that mean EC2 is already removing the volume, so it is not
+#: residual. Everything else -- ``available``, ``in-use``, ``creating``,
+#: ``error`` -- counts against the all-zero teardown gate.
+_VOLUME_TERMINAL_STATES = frozenset({"deleting", "deleted"})
+
 
 def _list_cluster_volumes(session: Any, region: str, project_name: str) -> list[str]:
     """Return EBS volumes tagged for a project cluster by the EKS CSI driver.
@@ -288,9 +293,16 @@ def _list_cluster_volumes(session: Any, region: str, project_name: str) -> list[
 
     Ownership is decided from the cluster name inside the tag key, which is the
     regional stack name, so a volume belonging to another project's cluster in
-    the same account is never claimed. State is deliberately not filtered: any
-    surviving volume tagged for a project cluster is residual, whether it is
-    ``available``, still detaching, or in ``error``.
+    the same account is never claimed.
+
+    Volumes EC2 is already removing (``deleting``/``deleted``) are not counted.
+    The cluster tag is not exclusive to CSI-provisioned PersistentVolumes: EKS
+    Auto Mode writes it onto node root volumes too, which live and die with
+    their instances. Those pass through ``deleting`` while a cluster tears
+    down, and a run that observed one mid-transition would fail the all-zero
+    gate for a volume that was already on its way out. Every other state --
+    ``available``, ``in-use``, ``creating``, ``error`` -- still counts, because
+    after teardown nothing should be holding a cluster-tagged volume at all.
 
     Enumerates unfiltered and matches client-side, like the other EC2 scanners
     here. A server-side ``tag-key`` filter would need wildcard semantics to
@@ -303,6 +315,8 @@ def _list_cluster_volumes(session: Any, region: str, project_name: str) -> list[
         volume_id = str(volume.get("VolumeId") or "")
         if not volume_id:
             raise RuntimeError(f"EC2 returned a volume without an ID in {region}")
+        if str(volume.get("State") or "") in _VOLUME_TERMINAL_STATES:
+            continue
         for key in _tags_to_dict(volume.get("Tags", [])):
             if not key.startswith(_CLUSTER_TAG_PREFIX):
                 continue
--- a/tests/test_floci_harness_inventory.py
+++ b/tests/test_floci_harness_inventory.py
@@ -178,6 +178,56 @@ class TestProjectResourceScanners:
         )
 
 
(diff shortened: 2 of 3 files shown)


Jmevorach:

Sanitized summary. Full reports and `checkpoint.json` stay local — they enumerate account identifiers, ARNs, and endpoint URLs.

- **Run ID:** `20260824T1705Z-df94acf1e902`
- **SHA:** `df94acf1e9027e7af2ded057ff21de2bcdbc9a1f`
- **Branch:** `fix/volume-scanner-ignore-deleting-state`
- **Profile:** `configured` (regional `us-east-1`; global / api-gateway / monitoring `us-east-2`)
- **Actions:** `--actions all` (12 selected, 12 passed, 0 failed)
- **Wall clock:** 2h 59m (21:04:51Z → 00:03:43Z)
- **Status:** **PASSED**, `fatal_error: none`

|---|---|---:|
| `preflight` | passed | 36.3s |
| `baseline` | passed | 564.2s |
| `deploy` | passed | 2518.0s |
| `topology` | passed | 710.6s |
| `api` | passed | 35.0s |
| `sqs` | passed | 54.9s |
| `central-queue` | passed | 25.5s |
| `schedulers` | passed | 48.1s |
| `opencost` | passed | 4.0s |
| `convergence` | passed | 21.3s |
| `destroy` | passed | 4836.9s |
| `final-inventory` | passed | 941.3s |

### The condition this patch exists for was actually present

Mid-teardown, the target Region held both kinds of volume under the same `kubernetes.io/cluster/gco-us-east-1` tag at once:

```
 5 GiB  available   <- stranded Alertmanager PV
50 GiB  available   <- stranded Prometheus PV
80 GiB  in-use      <- EKS Auto Mode node root
 4 GiB  in-use      <- EKS Auto Mode node root
80 GiB  in-use      <- EKS Auto Mode node root
 4 GiB  in-use      <- EKS Auto Mode node root
```

That is the shared-tag situation the previous run surfaced, and the reason the unfiltered scanner was a false-positive risk: node roots pass through `deleting` as the cluster tears down, and counting one mid-transition would have failed the all-zero gate for a volume already on its way out.

### Result

The sweep deleted only the two stranded PersistentVolumes, after the regional stack reached absence:

```
gco-us-east-1 is absent from CloudFormation.
Cleaned up 2 orphaned EBS volume(s) (55 GiB) left by cluster gco-us-east-1:
  - vol-… (5 GiB gp3, us-east-1a, pvc=alertmanager-…-db)
  - vol-… (50 GiB gp3, us-east-1b, pvc=prometheus-…-db)
```

Same 55 GiB split on fresh infrastructure as the v6.5.0 run — different volume IDs and a different AZ for the Prometheus volume, so the Prometheus/Alertmanager leak reproduces as a property of StatefulSet `volumeClaimTemplates`, not as an artifact of one deployment.

`final-inventory` then passed with `all_absent: true`, `residual: []`, `baseline_differences: []`, an all-zero residual summary, and **`cluster_volumes: 0`**. A post-teardown `describe-volumes` on the cluster tag returns **0**. Node root volumes went with their instances and were never counted or touched.

### Note on what this run can and cannot prove

It proves the gate still passes a genuinely clean teardown with the state filter in place, and it proves the runtime sweep continues to delete exactly the stranded volumes while leaving node roots alone. It does not prove the false positive would have fired without the patch — that depends on sampling a node volume during its `deleting` window, which neither run happened to hit. The Floci test pins that case deterministically from both sides: residual while the volume exists, not residual once EC2 has removed it.

PR #304, opened 2026-08-26
feat: pod-facing regional bucket ConfigMap and deployed job-policy read-back

Jmevorach:
## Summary

Two dependencies, both read/discovery surfaces over infrastructure that already exists — plus the refactor they turned out to need, the features that refactor made possible, and two bugs a live validation run found.

Sections 1 and 2 are the tickets. 3 through 6 are follow-on work from reviewing them and from running the result against a real account; each is a separate commit and reads independently.

**1. Expose the always-on regional bucket to job pods** (`feat(storage)`)

Every `GCORegionalStack` already provisions `gco-regional-shared-<account>-<region>` unconditionally, and already grants the job-pod role `s3:PutObject` on it. The gap was purely discovery: nothing told a pod the bucket's name. The three SSM discovery parameters are readable by the inference monitor but not by pods (the pod role carries no `ssm:GetParameter`), and no ConfigMap published the values.

This adds the `gco-regional-shared-bucket` ConfigMap to `gco-jobs`, `gco-system`, and `gco-inference`, mirroring the `gco-cluster-shared-bucket` pattern. **No IAM change** — the grant was already there.

Keys are `regionalBucketName` / `regionalBucketArn` / `regionalBucketRegion` rather than `sharedBucket*` so a pod can `envFrom` both ConfigMaps at once without either set clobbering the other: read central inputs, write local outputs. This makes the same-region bucket usable for training checkpoints, which previously had to either cross a region boundary to `Cluster_Shared_Bucket` or reconstruct the bucket name from account and region.

Unlike the cluster-shared placeholders, these are populated straight from the local constructs — no SSM round-trip, because this stack owns the bucket it is describing.

**2. Expose the deployed job validation policy** (`feat(api)`)

You could not find out whether a cluster would admit a job without submitting one. The only partial answer was `GET /api/v1/status`, which reported the cpu/memory/gpu caps, `allowed_namespaces`, and `validation_enabled` but omitted `allowed_kinds`, `trusted_registries`, `trusted_dockerhub_orgs`, the eight `manifest_security_policy` flags, `require_accelerator_toleration`, and `yaml_max_depth` — and nothing in the repo called it: no CLI command, no MCP tool.

Adds `GET /api/v1/policy`, surfaced as `gco jobs policy -r <region>` and the `get_job_validation_policy` MCP tool.

Two details worth reviewer attention:

- **Reading `cdk.json` is not a substitute**, for two independent reasons. It is the input to a deploy rather than the state of one, so it drifts the moment a stack is deployed from a different checkout. And `_augment_trusted_registries_with_project_ecr` appends the project's own ECR hostnames at synth time, so the effective registry allowlist is *always* strictly larger than the configured one. Two docs asserted the opposite — `RUNBOOKS.md` called `cdk.json` the "deployed-policy source of truth" — and are corrected here.
- **All three admission layers are reported**, not just the front door. The per-manifest caps are the first of three; a manifest can clear them and still be rejected by the namespace `LimitRange` (per container) or `ResourceQuota` (aggregate). Reporting only the caps would let a caller conclude a job is admissible when it is not. Layers two and three are read live from the Kubernetes API and degrade *per namespace* to `status="unavailable"` with a reason rather than failing the response or silently omitting the key, so a missing layer is always explicit.

Caps appear in the units the validator compares in (millicores, bytes) alongside the raw configured strings, because `"384"` vCPU and `384000` millicores are the same cap and a caller doing a local pre-check needs to know which it holds.

The route rides the existing `/api/v1/{proxy+}` greedy route, so no gateway change is needed to reach it per-region. It is added to `AGGREGATOR_REGIONAL_API_ROUTES` as a `GET` so the cross-region aggregator can read it too, matching that allowlist's read-mostly contract.

Also corrects `RUNBOOKS.md` to note that `gco jobs submit --dry-run` is a client-side `kubectl` parse, not an admission preview.

**3. The checks those two surfaces needed** (`refactor(admission)`, `feat(policy)`)

Reviewing (2) exposed the reason it could not be built on: every admission rule was a method reading `self`, so the only way to judge a manifest was to hold a live `ManifestProcessor` with Kubernetes clients attached. Anything holding a *policy* rather than a *processor* — an offline validator, a pre-submit check against a policy fetched over HTTP, a cross-region comparison — had no way in, and the only option left was to reimplement the rule. A second implementation of an admission rule drifts, and it drifts silently: the copy keeps passing while the real gate changes underneath it. `scripts/example_job_validation` was already importing `gco.services.manifest_processor`, and therefore `kubernetes`, to ask a question about a dict.

So `gco/job_admission.py` now holds the rules as functions of `(manifest, policy)` and nothing else, with no Kubernetes import. `ManifestProcessor` keeps all seven methods as thin delegates. Every moved name stays importable from `gco.services.manifest_processor` via explicit `X as X` re-exports, so none of the ~30 existing import sites change — that is what kept the risk down on a path that gates every job. Behavior is meant to be byte-identical, including messages and the two different valid-sentinels (`""` for caps, `None` for the others); the evidence is 672 pre-existing tests passing, not new ones.

On top of that, the questions `GET /api/v1/policy` enables but could not answer alone:

| Surface | Question it answers |
|---|---|
| `gco jobs check-policy MANIFEST [-r REGION]...` | Which regions would admit this, with per-region reasons |
| `gco jobs check-policy` (no manifest) | Do the regions still agree on policy |
| `gco jobs check-policy MANIFEST --offline` | Same evaluation from `cdk.json`, no AWS calls |
| `gco jobs submit --check-policy` | Advisory pre-submit check against the target region |
| `gco status --with-policy` | Cross-region agreement as a fleet section, each differing field a warn finding |
| `check_job_policy` (MCP) | The above, for an agent |

A 32-GPU job can be admissible in one region and over-cap in another, and until now you found that out by submitting. And because there are no per-region policy overrides, any field that differs across regions means a region was deployed from a different checkout — each region looks individually healthy, so that stays invisible until a manifest admitted yesterday is refused.

Three judgement calls worth reviewer attention:

- **Everything is advisory.** The cluster is the authoritative gate and this reads a snapshot of its policy over a network. A check that blocked on its own opinion would refuse valid jobs whenever it was stale or wrong. `--fail-on-reject` is opt-in, drift is `warn` so it does not trip `--fail-on-findings`, and a failed policy read never stops a submission that would have happened anyway.
- **An unreadable region is `unknown`, never `reject`.** It has not refused anything. Conflating the two would report a network failure as a policy violation and, under `--fail-on-reject`, fail a build over an undeployed region.
- **Registry drift is compared with ECR hostnames stripped.** CDK appends the project's own registry hostnames at synth time and those encode a region, so a raw comparison flags drift on every multi-region deployment — and a check that always fires is one people learn to ignore. The stripped entries are still reported separately as synth-time augmentation, since they are the concrete evidence that a locally-computed policy is not authoritative. That is also why `--offline` labels its own output as the *configured* rather than deployed policy.

**4. Two bugs the live runs found**

- `fix(rbac)`: `GET /api/v1/policy` returned `cluster_enforcement."gco-jobs" = {"status": "unavailable", "reason": "403 Forbidden"}` on a full validation run. `gco-manifest-processor-role` granted nothing on `resourcequotas` or `limitranges`, so layers 2 and 3 of the answer this PR added were simply unreadable and the endpoint could only ever report the caps. Granted `get`/`list` only, on a rule of its own so a verb added later for another resource cannot silently widen it.
- `fix(cli)`: `detect_container_runtime()` requires a runtime that *answers*, not one that is merely installed, and both call sites reported that single failure as "please install Docker, Finch, or Podman". That cost a full deploy attempt on a host with Finch installed and its VM stopped — the advice was to install software already present, so the actual fix (`finch vm start`) was not discoverable from the error. The message now branches on what is on `PATH`.

**5. A hole in the release-validation harness** (`test(live)`)

The run that surfaced the RBAC bug was green across all ten actions while the endpoint reported nothing usable. `cluster_resource_governance()` is fail-soft, so a Kubernetes read failure degrades that namespace and returns **HTTP 200**. `response.ok` is true, `call_api` succeeds, and every transport-level assertion in the harness passes. That is a hole, not just a miss: `/api/v1/policy` was this PR's own surface and the suite validated ten dimensions without touching it.

A new `policy` action asserts on the response *body*, per region and per namespace: all three layers present with `status == "ok"`; a `status: ok` must actually carry a `ResourceQuota` and `LimitRange` with a cpu/memory ceiling, since reporting "ok" with empty contents would satisfy a naive check while the ceiling went unenforced; identity (the body's region and `cluster_id` match the region addressed, `source` is the deployed runtime rather than a config file); and the project's own ECR hostnames appear in `trusted_registries`. That last one guards a feature nothing offline can: if the synth-time augmentation stopped, every job pulling a project-built image would be rejected as untrusted and no `cdk.json`-derived check would predict it. It runs after `topology` and before the mutating lifecycles, so a broken policy surface is reported before the suite spends time submitting jobs.

**6. Smaller follow-ons already on the branch**

| Commit | Change |
|---|---|
| `feat(storage)` | `gco storage s3-inventory` + `s3_inventory` MCP tool: every bucket the deployment owns, which are pod-writable, and how each is discovered |
| `refactor(capacity)` | Deleted the static GPU instance table; `instance_info` now resolves live from EC2 with `--region`. Fixed a real bug on the way: `gpu_memory_gib` was per-device on the API path, an 8x undercount on an 8-GPU instance |
| `chore(images)` | `APT_SECURITY_EPOCH` bump to pick up the openssl QUIC DoS fix (CVE-2026-14456) via `trixie-security`; no `.trivyignore` entry needed |
| `fix(mypy)` | `mypy_path = "gco_mcp"` so every invocation agrees with CI. The four `swarm_cmd.py` errors were a namespace-package resolution artifact — the `type: ignore` comments they flagged as unused are *required* under CI's command, and deleting them would have broken it |

## Type of change
- [x] `refactor:` No functional change

The two original changes are purely additive. The refactor is meant to be behavior-preserving and is backed by the pre-existing suites rather than new ones. The two fixes change behavior deliberately: the RBAC grant makes two previously-unreadable layers readable, and the runtime error message changes text that tests matched on. No existing resource is renamed or removed, and no existing policy value changes.

## Testing
- [x] `pytest tests/` passes locally
- [x] `cdk synth` succeeds (if CDK code changed)
- [x] If `examples/` changed: `gco examples validate --static-only` passes, and a live `gco examples validate --examples <changed>` run backs any behavior change — static: 379 checks, 0 failed. Live: `regional-shared-bucket-upload-job` submitted through its documented `cli-submit-direct` path and reached `complete` in 42s. Because the job reads all three ConfigMap keys with `os.environ[...]` and then calls `s3.put_object` with no exception handling, completion is end-to-end proof that the ConfigMap mounted, every key resolved, and the pod role's write grant works. Summary in a comment below.

New suites, all green:

| Suite | Result |
|---|---|
| `test_regional_shared_configmap.py` | 8 passed |
| `test_job_validation_policy_readback.py` | 19 passed |
| `test_storage_s3_inventory.py` | 26 passed |
| `test_job_admission_policy.py` | 26 passed |
| `test_job_policy_checks.py` | 37 passed |
| `test_status_policy_section.py` | 14 passed |
| `test_live_validation_policy.py` | 25 passed |
| `test_container_runtime_message.py` | 8 passed |
| `test_rbac_manifest.py` (2 added) | 49 passed |

Refactor safety, the number that matters most here: **672 pre-existing tests** across the full `manifest_processor`, security-validation, preservation, property, queue-processor and TrainJob suites pass unchanged against the extracted functions.

Also clean: `ruff check` and `ruff format --check` repo-wide; all three of CI's mypy invocations (`gco/ cli/ gco_mcp/ scripts/` at 269 files, `gco/stacks/ app.py`, and every `lambda/` subdir); `validate_k8s_manifests.py` at 172 manifests; `markdownlint` on every changed doc.

For the full suite I ran **CI's own shard selection** (`scripts/split_tests.py --shard 1 --of 2`, 179 files, 5330 passed) rather than a hand-picked list, after a hardcoded roster in the sibling example-validation suite broke CI once. Ten failures in that run were investigated individually: two are `-n auto` lock-timing flakes that pass serially, and the other eight reproduce identically on a pristine `origin/main` worktree, so they are pre-existing on this machine and not from these changes.

## Live release validation
- [x] Required and completed locally with `--actions all` — on `6d60e31`, the parent of the current head. The only difference is `0558f96`, a one-line `.gitignore` entry that cannot affect a deployment, so I did not spend another ~2.5h deploy/destroy cycle to re-attest it. The examples harness above *did* run against the current head.
- [ ] Not required (explain the applicability decision below)

**Applicability rationale:** this changes deployed infrastructure — a new ConfigMap applied to three namespaces on every regional cluster, an RBAC grant on the manifest-processor Role, a new IAM allowlist entry on the aggregator's identity and regional resource policies, and a new route served by the manifest-processor container.

A full `--actions all` run on this exact SHA passed **all 13 actions** across two regions, followed by teardown to zero residual resources — every `final-inventory` counter `0`, `baseline_differences` empty, all five stacks confirmed absent. Sanitized summaries are posted as comments below.

That run includes the new `policy` action, which reads the deployed policy back and confirms all three admission layers are readable: `cluster_enforcement."gco-jobs"` reports `status: ok` with a real `ResourceQuota` and `LimitRange`, where it reported `403 Forbidden` before the RBAC fix.

Two things that run established beyond the pass/fail:

- The deployed `trusted_registries` contained `<account>.dkr.ecr.us-east-1.amazonaws.com` and `<account>.dkr.ecr.us-east-2.amazonaws.com`, neither of which appears anywhere in `cdk.json`. That is the concrete evidence for this PR's central claim that reading `cdk.json` is not a substitute for reading the deployment.
- `gco storage s3-inventory` reported 10 known buckets, 8 deployed, with exactly the two regional-shared and cluster-shared buckets marked pod-writable.

## Checklist
- [x] Documentation updated (README, `docs/`, inline docstrings) as needed
- [x] `requirements-lock.txt` regenerated if `pyproject.toml` changed (n/a — no dependency change)

New docs: `docs/REGIONAL_SHARED_BUCKET.md` (with a bucket-selection table), an API reference section on reading the deployed policy, `gco jobs check-policy` and `gco status --with-policy` in `docs/CLI.md`, and a `policy` row in the live-validation contract table. Both new buckets' `RemovalPolicy.DESTROY` posture is called out explicitly in the checkpoint guidance, since a checkpoint bucket that gets emptied on teardown is a foot-gun worth naming.

MCP tool count moves 137 → 138 default (191 → 192 ceiling), updated in lockstep across the registry, the roster tests, and the three READMEs that quote it.


Jmevorach:
## Live release validation — full `--actions all` run

Ran locally against a dedicated development account, two regions (`us-east-1` workload, `us-east-2` global/api-gateway/monitoring). Account IDs, ARNs, and endpoint hostnames are redacted; the full report is not attached because it contains all three.

**Result: `passed` — all 12 actions.**

| Action | Result | Duration |
|---|---|---:|
| `preflight` | passed | 58s |
| `baseline` | passed | 552s |
| `deploy` | passed | 2679s |
| `topology` | passed | 432s |
| `api` | passed | 26s |
| `sqs` | passed | 45s |
| `central-queue` | passed | 36s |
| `schedulers` | passed | 48s |
| `opencost` | passed | 1s |
| `convergence` | passed | 21s |
| `destroy` | passed | 4835s |
| `final-inventory` | passed | — |

Teardown reached zero residual resources: every counter in the `final-inventory` summary is `0` (stacks, EKS clusters, S3 buckets, IAM roles, KMS keys, Lambda functions, load balancers, NAT gateways, ENIs, log groups, DynamoDB tables, SQS queues, ECR repositories, secrets, backup vaults and the rest), `baseline_differences` is empty, and all five stacks are confirmed absent from CloudFormation. Seven KMS keys are in the accepted pending-deletion state and two pre-existing CDK asset ECR repositories were correctly left alone as protected baseline.

### What this run established beyond pass/fail

**The deployed policy really does diverge from `cdk.json`.** This is the claim section 2 of the description rests on, and the run confirmed it concretely: the deployed `trusted_registries` contained

```
<account>.dkr.ecr.us-east-1.amazonaws.com
<account>.dkr.ecr.us-east-2.amazonaws.com
```

Neither string appears anywhere in `cdk.json`. `_augment_trusted_registries_with_project_ecr` adds them at synth time, so the effective allowlist is strictly larger than the configured one, and any tool that answers "will this be admitted" from the file alone will reject images the cluster would have accepted. That is why `gco jobs check-policy --offline` labels its own output as the *configured* policy and why its findings are advisory.

**`gco storage s3-inventory` against the live deployment:** 10 known buckets, 8 deployed, with exactly the cluster-shared and regional-shared buckets reported `pod_writable`. The two undeployed entries are the opt-in analytics buckets, correctly reported as not present rather than missing.

**`gco jobs policy -r us-east-1`** returned the full three-layer document against the deployed cluster.

### The bug this run found, and the harness gap behind it

On the first full run, layer 1 came back complete while layers 2 and 3 did not:

```json
"cluster_enforcement": {
  "gco-jobs": { "status": "unavailable", "reason": "403 Forbidden" }
}
```

`gco-manifest-processor-role` granted nothing on `resourcequotas` or `limitranges`. Fail-soft did its job — the response degraded per namespace instead of failing — but that also meant the endpoint could only ever report the front-door caps, which is exactly the "tells you a job is admissible when it is not" failure the description argues against. Fixed in `d5b420d` with `get`/`list` on a rule of its own.

Worth being explicit about the second-order problem: **all ten actions were green while this was broken.** `cluster_resource_governance()` degrades to HTTP 200, so `response.ok` is true and every transport-level assertion in the harness passes. `/api/v1/policy` was this PR's own surface and the suite validated ten dimensions without ever looking at its body. `c402129` adds a `policy` action that asserts per region and per namespace, placed before the mutating lifecycles so a broken policy surface is reported early.

### Not yet verified

This run could **not** confirm the RBAC fix end-to-end. The harness had no policy assertion at the time, and by the time I tried the endpoint by hand the regional API bridge was already deleted. The fix is covered by unit tests and the amended Role is present in the run's applied-manifest inventory, but it has not yet been *observed* returning `status: "ok"` from a live cluster.

A second full run on the current SHA is in progress for exactly that, and it will now fail loudly rather than silently if the enforcement layers are unreadable. I will post the result here.

Jmevorach:
## Live release validation — clean run on the final SHA

Two further runs since the summary above. Same dedicated development account and two-region topology; account IDs, ARNs, and endpoint hostnames redacted.

**Result: `passed` — all 13 actions, zero failures, on the SHA this PR currently points at.**

| Action | Result | Duration |
|---|---|---:|
| `preflight` | passed | 63s |
| `baseline` | passed | 555s |
| `deploy` | passed | 2440s |
| `topology` | passed | 534s |
| **`policy`** | **passed** | **1s** |
| `api` | passed | 25s |
| `sqs` | passed | 45s |
| `central-queue` | passed | 26s |
| `schedulers` | passed | 48s |
| `opencost` | passed | 1s |
| `convergence` | passed | 21s |
| `destroy` | passed | 4781s |
| `final-inventory` | passed | — |

`final-inventory`: every residual counter `0`, `baseline_differences` empty, all five stacks confirmed absent from CloudFormation, 7 KMS keys in the accepted pending-deletion state and 2 pre-existing CDK asset ECR repositories correctly preserved as protected baseline.

### The RBAC fix is now verified end to end

This is what the earlier run could not establish. Read back live from the deployed cluster:

```
us-east-1   cluster_id: gco-us-east-1   source: deployed-cluster-runtime
  caps: cpu=384000m  memory=4096Gi  gpu=16
  allowed_namespaces: ['gco-jobs']
  synth-time ECR hostnames in trusted_registries: 2

  cluster_enforcement:
    gco-jobs: status=ok
      ResourceQuota: ['gco-jobs-quota']
      LimitRange:    ['gco-jobs-limits']
```

That field was `{"status": "unavailable", "reason": "403 Forbidden"}` before `d5b420d`. All three admission layers are now readable, with real objects behind layers 2 and 3, and the new `policy` action asserts it rather than my asserting it.

The same action also confirms the synth-time ECR augmentation is live — two project ECR hostnames present in the deployed `trusted_registries` for this account, appearing in no `cdk.json`. Independent confirmation of the claim in section 2 of the description, and the reason `gco jobs check-policy --offline` reports its results as advisory.

### One transient failure, investigated rather than assumed

The first of these two runs failed `opencost` with `504 Gateway timeout` on `POST /api/v1/cost/reports` after 47.9s. It is not a regression, and the evidence is specific rather than a shrug:

- The identical action passed in **1.4s** on the previous run and **1.2s** on the run above.
- The only code difference between the failing and passing SHAs is the new read-only harness action and one test-file literal. Neither touches the cost pipeline or the manifest-processor runtime.
- The client gives mutating requests exactly one attempt by design, so a slow first report generation on a cold deployment fails the action outright rather than retrying.

I deliberately did not add a retry to make it pass. If that endpoint really cannot complete a first report inside the gateway's integration timeout under cold-start conditions, that is worth knowing as its own finding, and papering over it with a retry would hide it. It has not reproduced since. The failed run tore itself down cleanly through guaranteed cleanup, and `final-inventory` passed there too.

### Still open

The live examples harness for the new `examples/regional-shared-bucket-upload-job.yaml` has **not** run. It has its own deploy/destroy lifecycle, so it is a separate cycle rather than something foldable into the runs above. Static validation passes via `tests/test_example_job_validation.py`; the live half is the one unchecked box in this PR, and I have left it unchecked deliberately rather than implying it was validated.

That example is ticket C8's deliverable in executable form — a pod reading `regionalBucketName` from the new ConfigMap and writing to the bucket. Everything else in this PR has now been exercised against a real cluster; that path has only been exercised against a schema.

Jmevorach:
## Examples harness — the new example validated live

`python -m scripts.example_job_validation --examples regional-shared-bucket-upload-job`, against the same development account. This closes the last unchecked box.

**Static:** 379 checks, 0 failed (parse, transport gates, catalog symmetry, LimitRange/ResourceQuota fit for every example).

**Live:** `passed: 1, failed: 0, skipped: 0`

| Action | Result | Duration |
|---|---|---:|
| `preflight` | passed | 67s |
| `static` | passed | — |
| `baseline` | passed | 529s |
| `deploy` | passed | 2576s |
| **`examples`** | **passed** | **166s** |

The example itself, via its documented submission path:

```
name:        regional-shared-bucket-upload-job
submission:  cli-submit-direct  (gco jobs submit-direct)
region:      us-east-1
duration:    42.1s
criteria:    gco-jobs/regional-shared-bucket-upload -> complete
cleanup:     job.batch "regional-shared-bucket-upload" deleted from gco-jobs
```

### What "complete" actually proves here

Worth spelling out, because job completion is only as meaningful as what the job does. This one:

1. reads `regionalBucketName`, `regionalBucketRegion`, and `regionalBucketArn` via `os.environ[...]` — subscript access, not `.get()`, so a missing or empty key raises `KeyError` and the job fails;
2. those three come from `envFrom` → `configMapRef: gco-regional-shared-bucket`, the ConfigMap this PR adds;
3. then calls `s3.put_object` against the resolved bucket, with no exception handling, so a denied or failed write exits non-zero.

So a completed job is end-to-end evidence of the whole chain section 1 of the description describes: the ConfigMap exists in `gco-jobs`, `envFrom` mounts it, all three keys resolve to usable values, and the job-pod role's pre-existing `s3:PutObject` grant on the regional bucket actually works. A pod discovered the bucket without a hardcoded name and wrote to it.

That was the one claim in this PR still resting on a schema check rather than a live run.

### Note on a fix this run produced

The harness could not start with its documented default options. It defaults its report directory to `.example-job-validation/<run-id>` inside the repo, that path was not gitignored, and its own preflight requires a clean worktree — so it wrote a checkpoint and then failed on it:

```
[fail] preflight: Live validation requires a clean worktree; commit or remove these paths:
?? .example-job-validation/pr304-examples2/checkpoint.json
```

One line in `.gitignore` (`0558f96`). The sibling live-release-validation harness has had its `.live-release-validation/` entry all along; this one was missed when the examples harness landed. Unrelated to the two tickets, so happy to split it into its own PR if reviewers would rather keep this one scoped.

Teardown is running; I will confirm zero residual resources.

PR #326, opened 2026-09-02
feat: report the node and instance type a job's pod landed on

Jmevorach:
## Summary

The job read surface answered everything about a Job except the hardware it ran
on. That was harmless while a job pinned one exact instance type in its
manifest — the manifest *was* the answer. Once a Job is constrained to a *set*
of interchangeable instance types via a `nodeAffinity` `In: [...]` requirement,
[Karpenter](https://karpenter.sh/) chooses within that set, and the manifest can
only say what the run was **authorized** to use, not what it used.

Two things need the difference:

1. **Cost reconciliation.** Members of an eligible set differ in price, and a
   budget-ceiling estimate deliberately prices the most expensive one. Without
   the actual type, observed-vs-estimated cannot be closed.
2. **Diagnosis.** "Was this OOM on the smaller box?" is guesswork otherwise.

## What changed

`GET /api/v1/jobs/{ns}/{name}` gains a `scheduling` block:

```json
"scheduling": {
  "node_name": "ip-10-0-1-100.ec2.internal",
  "node_instance_type": "g5.2xlarge",
  "node_capacity_type": "spot",
  "node_labels": {"node.kubernetes.io/instance-type": "g5.2xlarge", "...": "..."},
  "nodes": [{"name": "...", "instance_type": "...", "capacity_type": "...", "labels": {}, "pods": [{"name": "...", "phase": "Running"}]}],
  "unscheduled_pods": 0,
  "node_lookup_error": null
}
```

- `node_name` comes from the pod's `spec.nodeName`; the instance and capacity
  types come from that node's `node.kubernetes.io/instance-type` and
  `karpenter.sh/capacity-type` labels, so spot-vs-on-demand landings are
  visible.
- The scalars describe the earliest-created scheduled pod, which is stable for
  the life of the job as retries add later pods. `nodes` covers every node
  involved, with the pods on each — a job that OOM'd on one member and was
  retried onto another reports both, with each pod's phase.
- `.../pods` carries the same block, additionally denormalized onto each pod as
  a `node` field so a caller iterating pods does not have to join.
- `JobInfo` carries the values through as top-level `node_name`,
  `node_instance_type`, `node_capacity_type`, `node_labels` and `nodes` fields,
  so `gco jobs get --output json` exposes them with no client change. The table
  output gains a Placement section and `gco jobs pods` gains an
  `INSTANCE TYPE` column.
- The TrainJob path resolves placement through its `<name>-node-0` child Job,
  the same convention the log fetch uses.

## Cost

One Node read per **distinct** node, on a path already talking to the cluster —
so exactly one extra API call for the common single-pod job. Confined to the
single-job reads: the list endpoint would pay it per job per region, so it does
not report placement at all (documented).

## Honesty over completeness

Nothing is inferred from the manifest. An unscheduled pod, a
garbage-collected pod (`ttlSecondsAfterFinished`), a node already reclaimed, or
a Node read refused for want of RBAC all report the instance type as **absent**,
with `node_lookup_error` explaining why in the last two cases. A node missing
the label reports `null`, not `"unknown"`, so a caller can tell absent from a
value. Resolving placement never fails the job read: a failure is recorded in
the payload instead. An unverified value that looks verified is worse than no
value.

Label values are also accepted only when they are genuinely strings, so a
malformed Node cannot put a non-serializable value into the response and turn a
successful job read into a 500.

## RBAC

Reading node labels needs core `nodes` `get`/`list`, added to the
`gco-manifest-processor-cluster-read` ClusterRole. Nodes are cluster-scoped, so
the rule cannot live in the namespaced `gco-manifest-processor-role`. Until that
manifest rolls out the feature degrades exactly as described above rather than
erroring — and `tests/test_rbac_manifest.py` now guards the grant so it cannot
be dropped, which would leave the feature permanently inert.

## Testing

`tests/test_job_node_placement.py` (47 cases) covers all three layers —
collector, routes, CLI client — pinning:

- the reported node, instance type and capacity type for a scheduled pod
- that a run on a non-cheapest eligible member reports **that** member
- a retried job that moved instance types reporting both nodes and phases
- one Node read per distinct node; one for a single-pod job; none on the list
  path; none for an unscheduled pod
- refused / 404 Node read, failed pod listing, missing label, malformed label
  values, label-less node, and pods with no creation timestamp
- the fields landing at the top level of the CLI's JSON payload, `JobInfo`
  mutable defaults not being shared between instances, and the extracted keys
  matching the `JobInfo` fields (guards the `**` splat against drift)
- the `Placement` table block, its absence when nothing is scheduled, and
  `gco jobs get` surviving a non-dataclass payload

Also run locally: `test_jobs.py`, `test_api_shared.py`, `test_manifest_api*.py`,
`test_rbac_manifest.py`, `test_cli_*`, `test_aws_client.py`, `test_output*`,
`test_mcp_server.py`, `test_gateway_route_coverage.py`, and the docs-coverage
guards — 1400+ passing. `ruff format`, `ruff check`, `mypy`, `yamllint` and
`markdownlint` clean. `docs/openapi/manifest-processor.json` regenerated.


Jmevorach:
## Verification status

### CI

All checks green on `0de5a347ce956a56072ad9c7f3962b16ba0eab16`: **85 pass, 0 fail, 2 skipped** (the two skips are the lockfile-freshness and fresh-install gates, which only run when dependency manifests change).

Two failures were found and fixed during review rather than waived:

- **`unit:pytest:core` (shard 2/3)** — the diagram artifact contract pins a content hash per charted source, so touching `cli/jobs.py` staled its three committed flowcharts. Regenerated incrementally (only `cli/jobs.py` re-rendered and restamped) in `5aafb53`.
- **CodeQL log injection, 4 alerts (Medium)** — the placement-failure warnings interpolate the namespace and job name, which arrive straight off the request path, and the Kubernetes error message echoes them back. Fixed in `0de5a34` by routing all three through `gco.services.structured_logging.sanitize_log_value`, the helper this repo already uses for untrusted values on the queue, manifest-processor, and job-admission paths. It neutralizes CR/LF, Unicode line separators, and other control characters, which is stricter than stripping `\r\n\t`.

  The unsanitized text still reaches `node_lookup_error` in the response body, where it is a JSON string that cannot forge anything and the caller needs the real reason. Both new guards assert on the arguments handed to the logger rather than a captured stream, because the service installs its own non-propagating structured handler — and each was confirmed to fail with the sanitizer stubbed out to a passthrough, so they are not vacuous.

### Live AWS release validation — clean terminal pass

Ran against the frozen SHA on a real multi-region deployment, deploying and destroying live infrastructure. A read-only baseline ran first and deployed nothing (`deployment_attempted: false`, preflight + baseline both passed), which is the precondition the launcher enforces before any destructive run may reuse it.

| | |
|---|---|
| Exact SHA | `0de5a347ce956a56072ad9c7f3962b16ba0eab16` |
| Branch | `feat/job-node-instance-type` |
| Requested actions | `all` (14) |
| Protected non-project stacks recorded | 5 |
| Inference matrix | vLLM + TGI, both digest-pinned, immutable model revisions |
| Process exit | `0` |

**14 of 14 actions passed, 0 failed:**

```
preflight        passed
baseline         passed
deploy           passed
topology         passed
inference        passed
policy           passed
api              passed
sqs              passed
central-queue    passed
schedulers       passed
opencost         passed
convergence      passed
destroy          passed
final-inventory  passed
```

Terminal state:

| Check | Result |
|---|---|
| Report status | `passed` |
| Cleanup completed | `true` |
| Checkpoint destroyed | `true` |
| Target stacks absent | `true` |
| Protected baseline unchanged | `true` (0 differences) |

Nothing was left standing and the five protected non-project CloudFormation stacks are byte-identical to the pre-run baseline.

### What this does and does not prove

The `deploy` action applied the RBAC change that grants the manifest-processor core `nodes` `get`/`list`, and `api` plus `convergence` exercise the deployed manifest-processor including the job read surface. So the endpoint shape, the RBAC grant, and the deployability of the change are all confirmed against live AWS.

What the live matrix does **not** isolate is the headline acceptance case — a job whose pod landed on a non-cheapest member of its eligible set. That needs a manifest with a multi-instance-type `nodeAffinity` requirement, so it rests on unit tests rather than this run. The degraded paths are likewise unit-tested, since they are failure injections rather than states a healthy deployment reaches: a Node read refused for want of RBAC, a node already reclaimed, and pods garbage-collected by `ttlSecondsAfterFinished`.

`tests/test_job_node_placement.py` carries 49 cases across all three layers (collector, routes, CLI client), including the cost bound (one Node read per *distinct* node, one for a single-pod job, none on the list path), a retried job that moved instance types reporting both nodes with the failing pod's phase, and the fields landing at the top level of the CLI's JSON payload.

All figures above are read from the run's sanitized summary. Account identifiers, ARNs, hostnames, and image digests stay in the owner-only `0600` log and are deliberately not reproduced here.

PR #333, opened 2026-09-03
test: enforce exact 100% Python and streaming-proxy JS coverage

Jmevorach:
## Summary

Raises both coverage gates from a 90/93 floor to **exact 100%**, after the suites actually reached it. At an exact floor there is no headroom, so a new uncovered line or branch fails on the pull request that introduces it rather than being absorbed by slack.

### Python — `gco/`, `cli/`, `gco_mcp/`

- 28 new behaviour-focused test modules close the last measured line and branch gaps (baseline was 93.16%: 2,111 missing statements, 1,524 missing branches).
- `fail_under = 100` in `pyproject.toml` `[tool.coverage.report]`, still the single source of the threshold. Shards keep `--cov-fail-under=0`; the floor is applied once to the combined data by `unit:pytest:core`.
- Where a branch was unreachable because of a producer or framework invariant, the branch was **removed**, not suppressed. No `pragma: no cover`, no `omit` additions, no denominator manipulation, and no mocks that contradict a real contract.

### JavaScript — `lambda/inference-streaming-proxy/index.mjs`

- `npm test` now scopes V8 coverage with `--test-coverage-include=index.mjs` and requires exact 100% lines, functions, and branches. The scope is the substantive part: the default also counted the test harness (`support.mjs`), so the previous number was not a statement about production code.
- Node's V8 reporter has no statement metric, so no statement figure is claimed anywhere in the docs.
- `unit:node:inference-streaming-proxy` tees the coverage table to an artifact under `set -o pipefail`, so the per-file numbers are reviewable without re-running the job.

### Frozen GitLab pipeline deprecated

Rather than maintain a coverage threshold in a file that never runs, `.github/legacy/` is removed along with every reference to it: the MCP `source://gco/config` allowlist entry and infra index line, the two guard tests that read it, the yamllint ignore, and the docs describing it. Two comments that claimed a live GitLab pipeline shares the semgrep rule list were corrected.

### Bug fixed along the way

Dropping the null parent/name guard in `stage_upload_path` during branch cleanup left `mypy --strict` with two `arg-type` errors — the guard had been doing double duty as type narrowing. The invariant (`_open_source` returns a null pair only for the root directory, which is opened `O_DIRECTORY` and so cannot be a regular file) is now stated with an `assert`: it narrows the types, fails loudly if broken, and adds no branch for coverage to miss.

## Testing

Local runs were targeted, never the full suite — CI is the authority on the combined number.

- `npm test` (Node 24.20.0 / npm 12.0.2): 93 tests, `index.mjs` at 100.00% lines / 100.00% branches / 100.00% functions, no uncovered lines.
- Targeted pytest across every touched area, plus the policy and documentation guards (`test_split_tests.py`, `test_ci_config_paths.py`, `test_docs_coverage.py`, `test_documentation_consistency.py`, `test_workflow_security_contract.py`, the MCP resource suites, `test_mcp_server.py`).
- `mypy gco/ cli/ gco_mcp/ scripts/ --exclude 'gco/stacks/'` and `mypy gco/stacks/ app.py`: clean.
- `ruff format --check` and `ruff check`, repo-wide `yamllint`, `actionlint`, `markdownlint-cli2` (0 issues across 125 files), `git diff --check`.
- `package-lock.json` and the generated `lambda/inference-streaming-proxy-build/` copy are untouched.

Rebased onto `main` after #331 and #332 landed.

## Checklist

- [x] Documentation updated (README, CONTRIBUTING, `.github/CI.md`, workflow READMEs, `docs/MAINTENANCE.md`, `tests/README.md`, wiki)
- [x] All 28 new test modules documented in `tests/README.md`
- [x] No coverage suppression used to reach the gate


Jmevorach:

Ran locally with `--actions all` against the approved validation account for this exact SHA.

| | |
|---|---|
| **Run ID** | `pr333-release-2b07059f-20260904T163038Z` |
| **SHA** | `2b07059f02806a56f5cb8c32c281fabc695cbfb5` |
| **Status** | `passed` (exit 0, no fatal error) |
| **Actions** | 14/14 passed, none skipped |
| **Duration** | 3h 18m |

### Per-action status

|---|---|---|
| preflight | passed | 32s |
| baseline | passed | 8m 27s |
| deploy | passed | 38m 28s |
| topology | passed | 12m 12s |
| inference | passed | 33m 35s |
| policy | passed | 3s |
| api | passed | 25s |
| sqs | passed | 55s |
| central-queue | passed | 36s |
| schedulers | passed | 1m 46s |
| opencost | passed | 5s |
| convergence | passed | 21s |
| destroy | passed | 1h 19m 42s |
| final-inventory | passed | 0s |

### Teardown and isolation

| Assertion | Result |
|---|---|
| `cleanup.completed` | `True` (cleanup was needed and ran) |
| `checkpoint.destroyed` | `True` |
| `final_inventory.stack_absence.all_absent` | `True` |
| `final_inventory.baseline_differences` | empty — no protected resource changed |

Independently re-checked after the run: no project CloudFormation stacks in either
deployment Region, no EKS cluster, no running instances, no Global Accelerator. The
protected non-project stacks captured at baseline were still present and untouched.

### Coverage of this change

`--optional-schedulers all`, inference in the regional Region with both runtimes:

| Runtime | Model | Image |
|---|---|---|
| vllm | `facebook/opt-125m` @ full 40-hex commit revision | digest-pinned (`@sha256:`) |
| tgi | `facebook/opt-125m` @ full 40-hex commit revision | digest-pinned (`@sha256:`) |

### Why this run matters for this PR

An earlier run on the pre-fix commit failed `topology`: the `health-monitor`
Deployment never reached minimum availability, because its `startupProbe` inherited
the kubelet's 1s `timeoutSeconds` default while the probe itself had to fork a
CPython interpreter and import `urllib` before issuing a request. The kubelet
reported `command timed out after 1s` and killed containers whose own logs said
`Application startup complete` a second later — 11 restarts across both replicas.

`topology` now passes on the fixed commit, and `inference` (which the failed run
never reached) passes too. That is the direct evidence for the probe-timeout fix in
this PR, not just a green re-run.

Full reports and `checkpoint.json` are retained locally with owner-only permissions
and are deliberately not attached: they carry account identifiers, ARNs, and endpoint
URLs.

PR #308, opened 2026-08-27
feat: encrypt ALB traffic to API workloads

Jmevorach:
## Summary
- re-encrypt ALB target traffic to TLS-only sidecars for the health monitor, manifest processor, and inference proxy
- bind application listeners to pod loopback, scope workload credentials away from the TLS sidecars, and hot-reload cert-manager leaves
- validate exact HTTPS/443 ALB listener and certificate state, HTTPS/8443 target groups, target health, certificate readiness, and generated Secrets

## Security model
API Gateway IAM authenticates original callers. The backend HMAC proves trusted-proxy key possession and request integrity on protected paths. ALB target TLS adds confidentiality for the final network hop; it is not mTLS because ALB does not validate target certificates.

## Compatibility
This intentionally removes plaintext API target listeners. The first protocol cutover on an existing deployment can be disruptive and should ship in the next major release rather than retaining an HTTP compatibility path.

## Validation
- exact head SHA: `e6bd1d72f13900bdca5729ccf3fb712d23eb6f32`
- focused local contract suite: 250 passed, 2 skipped
- focused credential/listener/rotation follow-up suite: 80 passed
- Ruff lint/format and Python compile checks passed
- GitHub Actions on the exact SHA: '84 passed, 3 expected skipped, 0 nonpass'
- fresh, untouched full live release validation: `tls-pr308-e6bd1d7-r10` passed all 13 actions; terminal cleanup completed and independent post-run AWS inventory is clean

## Type of change
- [x] Breaking change
- [x] Feature
- [x] Security hardening

PR #336, opened 2026-09-05
fix: queue-processor auth, applier phase guard, and deploy-time access config

Jmevorach:
Fixes the live SQS submission-path outage (queue processor unable to authenticate to the Kubernetes API after a hardening change split across the applier's base and post-Helm phases, crash-looping invisibly), plus three deploy-time configurability gaps and two CodeQL information-exposure alerts. One commit per workstream, reviewable independently.

## What changed, per workstream

**W2 — queue processor fails loudly (`fix(queue-processor)`)**
`load_k8s()` no longer falls through to `load_kube_config()` inside a pod: an in-cluster config failure raises a specific error naming the missing token path (`/var/run/secrets/kubernetes.io/serviceaccount/token`) and the ServiceAccount (new `SERVICE_ACCOUNT_NAME` downward-API env). `main()` distinguishes config failure from per-message failure: a config failure emits a structured terminal error, drains one message into a terminal FAILED job record (deleted only after the record is written; unparseable messages keep the DLQ path), and exits nonzero — so queue depth moves and the failure is visible instead of an infinite KEDA restart loop. On a message's final SQS delivery (`ApproximateReceiveCount` reaching the queue's `maxReceiveCount`), prevalidation/apply failures now write a FAILED record via the new `JobStore.record_job_failure` with error text bounded exactly like `central_queue_worker`. Earlier receives keep today's retain-for-retry semantics with no record. The ScaledJob gains `JOBS_TABLE_NAME` / `DYNAMODB_REGION` env (placeholders already existed in the stack's replacement dict).

**W1 — cross-phase consistency guard (`fix(applier)`)**
Planner invariant, enforced over both planned phases so the base pass fails before its first mutation: every planned workload running as a ServiceAccount that sets `automountServiceAccountToken: false` **and is RBAC-bound to the Kubernetes API** must mount a projected `serviceAccountToken` at the standard path. Violations surface through the existing `add_error`/`_planning_error_message` path naming the workload, the SA, and both source files. Upgraded clusters get the diagnostic the outage lacked: applying an automount flip onto a live cluster logs every planned workload running as that SA and whether each carries the projection. The real manifests directory passes; the synthetic outage shape fails planning (pinned in `TestAuthoritativeManifestPlanner`).

**W3 — developer EKS access surface (`feat(eks)`)**
`eks_cluster.developer_access` synthesizes one access entry per `{principal_arn, scope, namespaces}` (default: `AmazonEKSEditPolicy` namespace-scoped to `gco-jobs`; `scope: cluster` opts into cluster admin; absent config is purely additive; invalid entries fail synth). `eks_cluster.public_access_cidrs` restricts a public endpoint; enabling public access with no allowlist emits a loud synth warning naming the 0.0.0.0/0 exposure. `gco stacks eks endpoint set` is the audited config-only flip and refuses to widen access without an explicit `--cidr` list; `gco stacks status` reports configured-vs-live endpoint drift. `gco cluster doctor` reports reachability / authentication / authorization separately with per-case remedies, and distinguishes a stale kubeconfig context for a destroyed cluster ("no such host") from a private-only endpoint. Deploy output and the cdk.json comment now point at `gco stacks access` / `gco cluster tunnel` / `gco cluster doctor`; docs note the tunnel does not replace an access entry. MCP: `set_eks_endpoint_access` under `GCO_ENABLE_CONFIG_MANAGEMENT`.

**W4 — job-pod CloudWatch metrics (`feat(iam)`)**
One `cloudwatch:PutMetricData` statement on `service_account_role`, `Resource: *` constrained by `StringEquals` on `cloudwatch:namespace` (the action supports no resource-level scoping — same shape as every platform role's grant). Namespace configurable via `workload_metrics.cloudwatch_namespace`, defaulting to the GCO-owned `GCO/Workloads` rather than any one consumer's. cdk-nag acknowledgment mirrors the platform-role wording.

**W5 — regional shared bucket retention (`feat(storage)`)**
`regional_shared_bucket.removal_policy: destroy|retain` covering the bucket, its access-logs bucket, and its KMS key together (a retained bucket with a deleted key would be undecryptable). Invalid values fail synth. `gco storage s3-inventory` reports the effective policy (tolerant CLI-side read). Docs cover both modes and the interim copy-out path (`gco storage sync regional-shared:<region> ./checkpoints --prefix <prefix>`).

**W6 — CodeQL information exposure (`fix(api)`)**
All **ten** raw-exception 500 handlers in `gco/services/api_routes/jobs.py` (CodeQL flagged #308/#309; the same shape existed at eight more sites) now return a constant `"Internal server error"` detail. Full exception text stays in the server log via the existing `logger.error` calls; `from e` chaining intact; the intentional 400/404/409 and Kubernetes 502 branches unchanged. New test drives all ten handlers and asserts no exception text leaks. Alert #310 is a dismissed false positive — no change, as specified.

## Open questions, resolved

**Does `gco jobs submit-sqs` write a JobStore record for the job_id it enqueues?**
**No.** `JobManager.submit_job_sqs` (cli/jobs.py) generates the job_id and sends the SQS message; the job_id exists only inside the message. The feature request's fallback remedy ("create the record at submit time") is **unsafe as written**: the only record-creation path writes status `queued`, and the central queue worker claims QUEUED records for its region — a submit-time record would make the worker apply the job in parallel with the SQS processor (double apply), and the SQS message's `manifests` list doesn't fit the worker's single-`manifest` record shape either. Implemented instead: the processor creates the record itself, at failure time, directly in the terminal FAILED status via a conditional `attribute_not_exists(job_id)` put — a status the workers never claim, and a write that can never stomp a centralized-queue record. Successes remain unrecorded on the SQS path (unchanged); if submit-time records are wanted for successes too, that needs a lifecycle design for two writers and is out of scope here.

**W5 retention default: `destroy` (today's behavior), deliberately.**
Retain-by-default is the stronger data posture but silently changes teardown for every existing deployment and orphans a bucket + KMS key on every deploy/destroy validation cycle (each requiring manual cleanup before the region can redeploy under the same name). The knob makes retention a choice; the default keeps existing automation honest. Flagging loudly since the spec asked either way — no changelog file exists in the repo, so this PR description and the docs are the record.

## Spec statements found stale or refined (per the "say so and proceed" rule)

- **W2 IAM**: the spec asks to add `dynamodb:UpdateItem` to the manifest-processor role — it already grants `GetItem/PutItem/UpdateItem/Query/Scan` on the jobs table, and the `{{JOBS_TABLE_NAME}}`/`{{DYNAMODB_REGION}}` replacement wiring already exists. No stack change was needed for W2.
- **W1 invariant scope (refinement)**: the spec's literal rule ("every workload whose pod spec references that SA must project a token") fails the real manifests directory — the shipped inference-proxy and cost-monitor pods run hardened SAs with **no** Kubernetes API role and no projection, correctly. The implemented invariant scopes to SAs that are subjects of planned RoleBindings/ClusterRoleBindings (the inventory's own declaration that the SA talks to the API server). The outage shape is still caught; the acceptance criteria all hold.
- **CI trigger**: the spec says the draft PR triggers unit-tests/lint/integration-tests/security. Since #334, all four skip draft PRs; they run here via `workflow_dispatch` on the branch (as the spec's pre-PR fallback describes), and will run automatically when the PR leaves draft.

## Deliberately not done

- **Correlation id in 500 responses (W6 task 3, "propose rather than assume")**: proposed, not implemented. No request-id infrastructure exists anywhere in the API layer; the clean version is middleware assigning a request id, echoed in error details and bound into logs. Worth doing as its own change if wanted.
- **`endpoint_access` as an environment variable**: explicitly out of scope per the spec, and stayed out.
- **SQS-path success records**: see the W2 resolution above.

## Testing

- Per-workstream targeted test files locally (all green; new coverage in `test_queue_processor.py`, `test_template_store.py`, `test_kubectl_applier.py`, `test_workload_metrics_grant.py`, `test_regional_shared_bucket_retention.py`, `test_storage_s3_inventory.py`, `test_eks_access_config.py`, `test_cluster_doctor.py`, `test_stacks_eks_endpoint.py`, `test_managed_config.py`, `test_job_submission_and_service_api_edge_cases.py`).
- Full suites via GitHub Actions `workflow_dispatch` on this branch (results linked in checks/comments as they land).
- Live release validation (`gco release validate --actions all`) against the dedicated testing account: run status will be reported separately; reports stay local per protocol.


Jmevorach:
**Live release validation: PASSED** (run locally against the dedicated validation account; full reports stay on the operator machine per protocol)

- Run ID: `20260905T234953Z-d37037056db3`
- Commit: `d37037056db397c4b3955546a5ab80b9190fffa6` (this PR's head)
- Scope: complete `--actions all` run
- Overall report status: **PASSED**

| --- | --- | ---: |
| preflight | pass | 32s |
| baseline | pass | 598s |
| deploy | pass | 2772s |
| topology | pass | 495s |
| inference | pass | 2157s |
| policy | pass | 2s |
| api | pass | 24s |
| sqs | pass | 54s |
| central-queue | pass | 46s |
| schedulers | pass | 48s |
| opencost | pass | 2s |
| convergence | pass | 21s |
| destroy | pass | 4943s |
| final-inventory | pass | 540s |

Notes: the `sqs` action exercises the direct regional SQS job lifecycle end to end on this branch's queue-processor changes (W2) and the applier planner guard (W1). Off-by-default schedulers were reported per configuration (this branch touches no scheduler charts or the helm installer, so `--optional-schedulers` was not used). Zero residual resources; baselines preserved exactly.

PR #338, opened 2026-09-06
feat: correlate generic API 500s with request ids

Jmevorach:
Follow-up proposed in #336 (W6 task 3, "propose rather than assume"): the jobs routes' 500 handlers now return a constant detail with no exception text, which leaves a client-reported failure with nothing to grep for — the operator has to match timestamps against the service log. This PR gives every manifest API request a server-generated correlation id that appears in three places at once, so a reported id greps directly to the logged exception:

- the `X-Request-ID` response header, on every response (success or error);
- the generic 500 detail — `Internal server error (request-id: <32-hex>)`;
- the paired server-side error log line.

## What changed

- **`gco/services/request_context.py`** (new): a `ContextVar`-backed correlation id, bound per request by new middleware on the manifest API app. Ids are **always generated server-side** (`uuid4().hex`, log-safe by construction); an inbound `X-Request-ID` header is deliberately ignored — a client-controlled value adjacent to log lines would need CWE-117 sanitization and could be replayed to muddy an investigation. `current_request_id()` binds on first use, so the log line and the response detail produced in the same context always agree even when no middleware ran (direct route invocation in tests).
- **`api_shared.internal_server_error(context, error)`** (new shared helper): logs the full exception with the id, returns the generic correlatable 500. Adopted at all **fifteen** handler sites across the manifest API routers: the ten `jobs.py` handlers hardened in #336, plus `manifests.py` (4) and `templates.py` (1). Four of those five were still embedding raw exception text in their 500 details — the same information-exposure shape CodeQL flagged on `jobs.py` — and are now generic + correlatable.
- The DEBUG-gated global exception handler reports the id in its JSON body (`request_id` field) and in its log line.
- `docs/API.md` documents the header and the "report this id" flow.

## Testing

- `tests/test_request_context.py` (new): id format/uniqueness, bind/unbind context hygiene, bind-on-first-use stability, context isolation, the 500-helper detail/log pairing (asserts the marker exception text appears in the log and never in the detail), and the global handler's `request_id` field.
- `tests/test_manifest_api.py`: middleware pinned end-to-end over the real app — unique 32-hex `X-Request-ID` per response, and an inbound header is never echoed back.
- The #336 leak test now asserts the exact new detail shape (`Internal server error (request-id: <hex>)`) across all ten jobs handlers, still proving no exception text leaks.
- `request_context.py` at 100% line+branch coverage locally; full suites via workflow_dispatch on this branch (drafts skip PR-event CI).

## Scope notes

- `health_api.py` and `inference_api.py` are separate services with their own apps; wiring them up would be a copy of this pattern and is left out deliberately.
- The two-commit shape is the diagram-catalogue contract: source change first, then the pinned regeneration commit.


Jmevorach:
## Live release validation: PASSED ✅

- **Run ID:** `20260906T041715Z-db830a5f12cb`
- **Commit:** `db830a5f12cb7b551d3a5512cad83a8f6821e98f` (current PR head)
- **Action scope:** `all` (14/14 actions)
- **Overall status:** `passed` — zero residual resources at final inventory, exact baseline preserved

|---|---:|---:|
| `preflight` | passed | 27.730s |
| `baseline` | passed | 542.805s |
| `deploy` | passed | 2548.051s |
| `topology` | passed | 580.590s |
| `inference` | passed | 2111.446s |
| `policy` | passed | 2.195s |
| `api` | passed | 24.623s |
| `sqs` | passed | 44.667s |
| `central-queue` | passed | 25.248s |
| `schedulers` | passed | 47.681s |
| `opencost` | passed | 2.040s |
| `convergence` | passed | 20.885s |
| `destroy` | passed | 4998.781s |
| `final-inventory` | passed | 0.000s |

Notes:
- The `api` action exercised the deployed manifest API with the request-correlation middleware in the hot path (authenticated job lifecycle against the live endpoint).
- `final-inventory` duration is a report-recording quirk (console measured ~538s); the action ran fully and passed.
- Extra live-path coverage beyond the required suite: `Floci Tests` (including `floci:e2e:release-validate` emulator harness) dispatched at this SHA and passed (run 34010740198).
- Full reports retained locally per docs/LIVE_RELEASE_VALIDATION.md (not committed).

PR #375, opened 2026-09-11
platform services: one hosting contract, enforced network posture

Jmevorach:
## Summary

A review of how the GCO platform services are hosted on the cluster — the five `gco-system` Deployments (health-monitor, manifest-processor, inference-monitor, inference-proxy, cost-monitor), the KEDA `ScaledJob` queue processor and the TLS sidecar — with the fixes applied in the same PR, and the network posture those services and the jobs run under brought in line with what the docs already promised.

The question that started it, "should HPA be on by default?", gets a per-service answer below. The review found two correctness bugs in how replicas interact with the code (every webhook delivered once per health-monitor replica; the webhook table name wrong for any non-default `project_name`), a disruption budget that let 8 of 10 inference-proxy pods be evicted at once, one Deployment that followed none of the conventions the other four do, `cdk.json` sizing knobs that were validated as required and then ignored — and, on the network side, a set of NetworkPolicies that EKS Auto Mode was accepting and enforcing with nothing, because the controller that enforces them is off until a `kube-system/amazon-vpc-cni` ConfigMap turns it on. With enforcement actually on, several of the shipped rules would have broken real traffic (DNS from any pod, Prometheus scraping the inference-monitor, a multi-pod job talking to itself, jobs reaching S3/DynamoDB/STS), so the rules were corrected in the same change and the kind e2e now proves the verdicts both ways.

Seven full live release validation runs followed. Each of the first six stopped on a different finding — a reusable S3 bucket name colliding with its own deleted predecessor, NetworkPolicy semantics on the VPC CNI (Service ClusterIPs, operator-to-workload paths, the policy attach window), a replayed HMAC envelope on retry, a Slurm token signed with a stripped key, and health-monitor liveness probes killing a healthy container on a CPU-starved node — and every one was fixed in this PR rather than retried; the seventh passed all sixteen actions. The follow-up section below records each finding, the fix, and the alternatives rejected.

## Type of change

- [x] `ci:` CI / tooling change

## How the services were hosted before this PR (evidence in the manifests, applier and stack)

| Service | Replicas | Strategy | Autoscaling | PDB | Spread | Rolls on every deploy | Background work per replica |
|---|---|---|---|---|---|---|---|
| health-monitor | 2 | RollingUpdate 0/1 | — | minAvailable 1 | soft zone+host | yes | poll loop, CloudWatch publish and the **webhook dispatcher in every replica**; only the ALB→SSM write was Lease-elected |
| manifest-processor | 3 (hard-coded; `cdk.json` `replicas` ignored) | RollingUpdate 0/1 | — | minAvailable 2 | soft | yes | central queue worker in every replica (safe: DynamoDB conditional claims) |
| inference-monitor | 2 | *unset* (25%/25%) | — | minAvailable 1 | none | **no** | Lease-elected leader; others are standbys |
| inference-proxy | 3 | RollingUpdate 0/1 | HPA 3–10 (cpu/mem + TLS cpu) | **minAvailable 2** | soft | yes | stateless |
| cost-monitor | 1 | Recreate | — | none | none | yes | idempotent report writer (single writer by design) |
| queue-processor | KEDA ScaledJob 0–10 | — | queue depth | — | — | — | one SQS message per Job |

The applier creates or patches each object (no server-side apply), re-asserts `spec.replicas` on every convergence unless the Deployment carries `gco.aws/hpa-controls-replicas: "true"`, never waits for rollout itself, and a later Step Functions task validates every object's readiness. Images are CDK `DockerImageAsset`s referenced by their content-hash tag from the bootstrap assets repository; `metrics-server` is an EKS add-on, so HPAs have a signal.

## Should HPA be on by default? Per service

- **inference-proxy — yes, and it is.** Stateless, per-request work, the one service whose load scales with traffic. Kept; min/max become `cdk.json` knobs (`inference_proxy.min_replicas` / `max_replicas`, defaults 3/10) instead of manifest literals.
- **manifest-processor — not by default; available as an opt-in.** It is I/O-bound (the API handlers block the event loop on Kubernetes round trips), so CPU is a weak saturation signal, and each extra replica also runs the central queue worker, which adds pollers competing for the same DynamoDB claims rather than throughput. The right first step is to make the size an operator decision: `manifest_processor.replicas` and `resource_limits` — already validated as required in `cdk.json`, never wired — now render into the manifest (limits floored at the manifest's 500m/1Gi requests), and `manifest_processor.autoscaling` (`enabled: false` by default, `max_replicas`, `cpu_target_utilization_percentage`) ships a CPU HPA (`35-manifest-processor-hpa.yaml`) for deployments that want it, using the same `hpa-controls-replicas` mechanism the proxy uses. The applier prunes the HPA again when it is turned back off.
- **health-monitor — no.** Its work is a per-replica poll of the cluster plus CloudWatch publication; more replicas mean more identical API load, not more capacity, and the API it serves answers from a cache. Two replicas are for availability. What *was* wrong: the webhook dispatcher ran in both replicas with no coordination, so every job transition fired every webhook twice. It is now Lease-elected (`gco-health-monitor-webhooks`, pre-created like the ALB-sync Lease so RBAC stays `get`/`update` on one named object) through a shared `gco/services/leader_lease.py`, with at-most-once delivery across a failover window bounded by the lease duration.
- **inference-monitor — no.** Leader-elected controller; extra replicas are hot standbys. Two stay.
- **cost-monitor — no.** Single writer by design (deterministic S3 keys, `head_object` skip); `Recreate` with one replica is the honest shape and the `karpenter.sh/do-not-disrupt` annotation protects it from consolidation.
- **queue-processor — already autoscaled by KEDA** on queue depth; that is the right signal for a one-message-per-Job worker.

## Findings fixed — hosting

**Disruption budgets.** `minAvailable: 2` on a Deployment the HPA can take to 10 replicas allowed eight simultaneous evictions of the inference data plane. Every multi-replica service now uses `maxUnavailable: 1`, which scales with the replica count and works for any operator-chosen size; the readiness validator still requires `currentHealthy >= desiredHealthy`. Every Deployment also carries `minReadySeconds: 10`, so a pod that passes readiness and dies immediately does not advance a rollout.

**inference-monitor followed none of the conventions**: no rolling-update strategy, no deployment-timestamp annotation (so a config-only deploy never rolled it), no `imagePullPolicy`, no startup probe, liveness and readiness that only asserted `/proc/1` exists (a check that cannot fail while the container runs), no topology spread or anti-affinity, an automounted API token where the other RBAC-bound services project one explicitly, no termination grace tuned to its reconcile loop. It now matches: RollingUpdate 0/1, timestamp, pull policy, real probes against the Prometheus endpoint it already serves on :9090, spread and anti-affinity, `automountServiceAccountToken: false` plus the projected token pair, `terminationGracePeriodSeconds: 60`, and a `seconds_since_last_pass` gauge so a stuck loop is visible.

**Configuration that lied.** `manifest_processor.replicas` and `resource_limits` were required by the validator and read by nothing; they now drive the manifest. Seven `*_THRESHOLD` / `PENDING_*` variables were injected into manifest-processor and read by no code; removed. The health-monitor omitted `WEBHOOKS_TABLE_NAME` and `ALLOWED_NAMESPACES`, so its dispatcher fell back to the literal `gco-webhooks` table (wrong for any other `project_name`) and to `gco-jobs,default` instead of the configured namespace policy; both are now injected from the same placeholders the manifest-processor uses.

**Consistency and hygiene across all five Deployments**: `imagePullPolicy: IfNotPresent` (the tags are content hashes, so a cached layer is the right layer, and a pod restart no longer depends on ECR being reachable); `revisionHistoryLimit: 3`; `enableServiceLinks: false` (no `COST_MONITOR_PORT`-style env injection into unrelated pods); `sizeLimit` on every `/tmp` emptyDir; uvicorn `timeout_graceful_shutdown` honoured from `GRACEFUL_SHUTDOWN_TIMEOUT_SECONDS` by all four FastAPI services, not just the proxy; cost-monitor probes moved to exec-on-loopback so the only ingress rule it needs is the manifest-processor one.

**Stack.** The three `ecr.Repository` constructs the stack created, scanned and never used are gone. Every platform ServiceAccount that carries an IRSA annotation now also has an EKS Pod Identity association (inference-monitor and cost-monitor were missing theirs), and a dangling association for a `gco-system/gco-service-account` that no manifest creates is dropped; a test pins IRSA↔Pod Identity in both directions.

## Findings fixed — network posture

**Enforcement was off.** EKS Auto Mode ships the network policy agent in its node networking stack, but the controller that turns NetworkPolicy objects into enforced rules stays off until `kube-system/amazon-vpc-cni` says `enable-network-policy-controller: "true"` (AWS docs, "Use Network Policies with EKS Auto Mode"). GCO never applied that ConfigMap, so every policy in `03-network-policies.yaml` was accepted by the API server and enforced by nothing. New `06-network-policy-controller.yaml` applies it, rendered from `eks_cluster.network_policy_enforcement` (default `true`; `false` is the documented escape hatch, and the CUSTOMIZATION doc explains why adding a NetworkPolicy is the intended fix instead).

**Rules that would have broken traffic once enforced**, corrected in the same change:

- `allow-dns` admitted DNS only to pods in `kube-system`, which pins the rule to one resolver topology; on Auto Mode the resolver path is AWS-managed and not something a `podSelector` should have to name. DNS (53 UDP/TCP) is now allowed to any destination in all three namespaces — the port is the property worth guarding, not the resolver's address.
- Nothing admitted the inference-monitor's `:9090` metrics port, so its new probes and the Prometheus scrape would both have failed. A source-less `allow-metrics-to-inference-monitor` admits exactly that port, in the same shape as the ALB→8443 rules (the kubelet is not a pod, so a `from:` clause cannot name it).
- `gco-jobs` had default-deny ingress with an allow only between pods labelled `ray.io/cluster: ray-cluster`, and egress only to kube-system DNS and to port 443 inside the VPC CIDR blocks — so any multi-pod job outside Ray could not reach its own peers, and once enforced no job could reach S3, DynamoDB, STS (no VPC endpoints existed, so those calls left through NAT), Hugging Face, or pip. The namespace now gets `allow-same-namespace` (ingress and egress within the namespace), `allow-https-egress` (443 anywhere), and `allow-vpc-egress` (any port to the VPC CIDR blocks, so in-cluster services and the new gateway endpoints work); cross-namespace ingress stays denied.
- The retired `allow-vpc-endpoint-egress` and `allow-ray-cluster-internal` policies are swept from existing clusters through the applier's `_LEGACY_REMOVED_RESOURCES`.

**VPC endpoints.** `vpc_endpoints: {gateway: [s3, dynamodb], interface: []}` in `cdk.json` creates S3 and DynamoDB gateway endpoints in every regional VPC by default (free, and they take model/dataset/checkpoint/cost-report traffic off the NAT gateways' per-GB metering); interface endpoints (PrivateLink, billed per AZ-hour) are opt-in by validated service name. Documented in CUSTOMIZATION ("VPC Endpoints"), with the ARCHITECTURE "Network Security" table rewritten to describe the posture that is now actually enforced.

## Follow-up after the first live runs: bucket names, enforcement fixes, diagnostics, harness, Floci

Four full live runs of this branch each stopped on a different finding, and every finding was fixed rather than retried. The first failed at `deploy`: `gco-monitoring` rolled back because S3 answered the deterministic `gco-cost-reports-<account>-<region>` name with `409 BucketAlreadyExists` minutes after the previous run had deleted that very bucket — deleted bucket names are held for an unspecified window, so any deploy that follows a destroy of the same project in the same account and Region can collide with itself. The run's guaranteed cleanup then finished with every stack gone, yet `final-inventory` went red on two Tagging-API records of gateway VPC endpoints EC2 had already deleted. The second run deployed cleanly and failed at `inference`: the enforced NetworkPolicies were denying the inference proxy's connections to a healthy model endpoint. The third got past inference and failed at `schedulers` because the same default-deny cut the Slinky operator off from the Slurm REST API. The fourth reached the Slurm probe itself, which signed its token with a key it had altered. The fifth passed every workload action and failed the last one, `network-posture`, on the one probe whose verdict depends on the client pod's own policies: the client dialed before the VPC CNI had attached them. The sixth failed the `topology` check on an unhealthy ALB target: a health-monitor that was answering every request was being restarted by its own liveness probe on a CPU-starved node.

- **Every S3 bucket is now CloudFormation-named** ([ADR-0005](docs/adr/0005-cloudformation-generated-s3-bucket-names.md)). The cost-report, cluster-shared, regional-shared and analytics-studio buckets drop their explicit `bucket_name`; the monitoring stack publishes `/<project>/cost-report-bucket/{name,arn,region}` and grants the regional cost-monitor roles through its bucket policy and KMS key policy (nothing in the regional stacks names the bucket any more; they hold `ssm:GetParameter` on exactly that parameter). The cost-monitor service discovers the bucket lazily through `CostReportBucketLocator` (skips its scheduled pass with a reason, and the API answers 503, until the parameter is published; refreshes after a TTL; keeps the last name over a transient SSM failure), `gco storage` and the OpenCost release check read the same parameter, and a static naming-contract test fails if a `bucket_name=` creeps back into the stacks. Rejected alternatives (retry-on-409, a random suffix under a predictable prefix, moving the bucket to the global stack) are recorded in the ADR.
  **Upgrade consequence: updating an existing deployment replaces these four buckets.** CloudFormation treats the name change as a replacement, so their contents must be copied out first; the ADR spells this out.
- **NetworkPolicy enforcement and Service ClusterIPs (found by the second run).** The VPC CNI evaluates a pod's egress before kube-proxy translates a ClusterIP, and the network policy controller adds a Service's ClusterIP to an egress rule only when the Service's own `spec.selector` matches the rule's peer `podSelector`. `allow-inference-proxy-to-inference` selects `gco.io/type: inference`, but the inference Services selected on their per-endpoint `app` label alone, so the pods were admitted and the Service in front of them was not: the kubelet kept getting 200 from the model pod while every proxy request timed out. Every ClusterIP Service the inference monitor creates (endpoint, PD proxy, per-role) now carries `gco.io/type: inference` in its selector; a manifest test pins the alignment between Service selectors and policy peers; ARCHITECTURE and CUSTOMIZATION explain the property for the next rule anyone writes; and the cost-monitor Service listens on its container port (8080) so the manifest-processor path does not depend on the controller's port mapping either. Calico in the kind job does not have this limitation, which is why the kind probes passed. **Upgrade note:** managed endpoints created before this change keep their old Service selector until re-deployed.
- **Streaming proxy retries replayed a single-use envelope (found by the same run).** The inference streaming proxy retries GET/HEAD on 429/502/503/504 but re-sent the first attempt's HMAC envelope; the backend refuses a repeated nonce with 403, so every retryable failure surfaced as "requests must come through authenticated API Gateway". Each attempt is now signed afresh (Node tests cover it at 100%).
- **Operators that drive their workloads through the workload's own API (found by the third run).** The Slinky operator runs in its own namespace and reconciles NodeSets through slurmrestd (node and reservation state, draining workers before their pods go); with `gco-jobs` default-denying ingress from other namespaces every reconcile timed out on port 6820, the workers never converged, and the smoke job sat in `PENDING` for its whole budget. `allow-slurm-operator-to-restapi` admits the operator's namespace on the REST port only and is pruned with the other Slurm-gated rules when Slurm is disabled. The finding generalizes to KubeRay, the other operator GCO installs that polls its workloads directly: the RayJob and RayService controllers read job status and Serve health from the head's dashboard on 8265, so `allow-kuberay-operator-to-ray-head` admits the operator's namespace to head pods on that port. Tests pin both namespaces to the ones `charts.yaml` installs the operators into, and the base manifest's "no cross-namespace ingress in `gco-jobs`" test names the KubeRay rule as its single exception (the Slurm rule ships in the Slurm-gated manifest with its own test). The kind job runs neither operator and earlier live runs never enforced the policies, which is why nothing caught this before.
- **The Slurm probe signed with an altered key (found by the fourth run).** slurmrestd answered `511` while slurmctld logged `cred_verify: jwt_decode failure: Invalid argument`. slurmctld mmaps the HS256 key file and verifies against every byte of it, and the Slinky chart generates that file with `randAscii`, whose alphabet includes the space character; both the harness probe and `examples/slurm-cluster-job.yaml` stripped the key before signing, so on the roughly one cluster in fifty whose key begins or ends with a space they signed with a different key than the controller held. Both now use the mounted bytes unchanged, and the manifest test pins the raw read.
- **A network probe's verdict is its steady state (found by the fifth run).** The VPC CNI attaches a new pod's policies in parallel with the pod's start and admits all of its traffic until they are in place (standard mode; Auto Mode's NodeClass `networkPolicy: DefaultDeny` is the strict alternative and needs a policy for every pod on the node, so GCO does not set it). The port-80 egress probe dialed once as its first instruction and read that window instead of the policy; the other five probes were unaffected because their deny verdicts come from the long-running listeners' ingress rules, and Calico in the kind job attaches policies before the pod starts. Each probe now samples until one answer has held for 30 seconds (no earlier than 45 seconds in), exits 43 rather than guessing when the answer keeps changing, and prints every change of answer; the harness keeps those as `samples`, names an unsettled probe in the failure, and flags `attach_window_observed` so the window stays visible as evidence without deciding the action. ARCHITECTURE records the timing property next to the ClusterIP one.
- **Exec probes start lean and liveness gets a minute (found by the sixth run).** The node hosting the health-monitor leader was spending most of both vCPUs in the kernel (the node's logs do not name the cause; the other nodes sat at 2–18%), and on it the container's exec probes — `python -c "import urllib.request; …"`, ~130 modules — timed out 190 times in 25 minutes at 3 s and 5 s; liveness restarted the container five times while its log showed it answering everything that reached it, and the ALB marked the pod unhealthy. The other API containers on the same node timed out twice each: the health-monitor had the smallest CPU request of the three, so under contention it was starved first, and a probe is a process competing with the server for the container's quota. Every exec probe now starts the interpreter with `-I -S` and speaks HTTP over a bare socket with bounded timeouts (about a fifth of the work per probe; the kubelet cannot reach a loopback listener itself, so the fork stays but gets cheap), liveness restarts only after a minute of continuous failure (4 × 15 s, 10 s per attempt) instead of 45 s, startup gets the same 10 s, readiness 5 s, and the health-monitor's request rises to 200m, its leader's measured usage. The probe-timing contract pins the lean prefix, the inner socket timeouts staying below the probe timeout, the 10 s liveness floor and the minute-long restart window.
- **Deploy diagnostics name the resource that failed.** `StackManager._diagnose_deploy_failure` walks the stack events back to the operation's "User Initiated" start and prints each root-cause `*_FAILED` event with its logical id, type and reason (skipping the "Resource creation cancelled" cascade) before the stack verdict and the existing advice — the 409 above had to be dug out of the event stream by hand.
- **Harness hardening.** Preflight refuses a run that will deploy unless the checkout, the report directory and the home volume each have 20 GiB free (`--min-free-disk-gib`; `0` disables); the deploy action prunes the `cdkasset-*` images, the mirrored CDK asset repositories and dangling layers it produced, best effort, with the evidence in the deploy record. Tagging-API records of VPC endpoints are stripped from `baseline` and `final-inventory` only after EC2 itself answers `InvalidVpcEndpointId.NotFound` or reports a terminal state, recorded as `accepted_deleted_vpc_endpoints` evidence (the same posture as the deleted-table DynamoDB streams); the second run's final inventory accepted an endpoint record the index was still serving eight hours after its stack had gone.
- **Floci coverage for the paths the unit suite could only fabricate.** Six new emulator modules run the production code against real emulated services: the bucket-identity contract end to end (a real CloudFormation stack with an unnamed bucket and the three parameters; the locator, the release check and `gco storage` all land on the bucket CloudFormation created; real `ParameterNotFound` and S3 404 on the negative paths), the full `gco storage` inventory over all four owning stacks, the endpoint acceptance against a live-then-deleted gateway endpoint, the deploy-failure diagnostics and stuck-stack pre-check on a stack that really rolls back on a bucket-name collision, the Mission `DynamoDBBackend`, and `CapacityHistoryStore`.
  **Bug found by the new layer and fixed:** the Mission `DynamoDBBackend` (excluded from unit coverage, never smoke-tested) could not save a single real session — boto3's resource API rejects `float`, and every session carries floats in criterion targets and observed metrics. Numbers now cross the wire through the same Decimal conversions the capacity-history and mission-memory stores use, so a session loaded from DynamoDB is indistinguishable from one loaded from the filesystem backend.
- **Image scans.** Trivy began failing the six service images on CVE-2026-11822 and CVE-2026-11824 in `libsqlite3-0` mid-branch; `APT_SECURITY_EPOCH` moves to `2026-09-12` across the seven Dockerfiles so the rebuilt images pull the fixed package.

Observation carried in the validation report rather than fixed in code: in the fourth run the `topology` action took 45 minutes because Grafana's first container sat for 42 minutes after "Starting Grafana" before the kubelet restarted it, after which the Helm-release checks passed on the first try; nothing in the platform's configuration changed between that run and the ones before it, so it is recorded as an environmental stall, not a regression.

## Testing and CI added

- `tests/test_platform_workload_contract.py` (70 tests) — one contract for every `gco-system` workload derived from the manifests: resources on every container, startup/liveness/readiness, the full securityContext, rolling-update shape, pull policy, deployment-timestamp, `automountServiceAccountToken: false`, bounded emptyDirs, spread + anti-affinity and a `maxUnavailable: 1` PDB for every multi-replica Deployment, `minReadySeconds`, HPA ↔ `hpa-controls-replicas` pairing in both directions (including the placeholder-gated manifest-processor HPA), every workload selected by an ingress allow, Service targetPorts, and the applier's `PLATFORM_DEPLOYMENTS` equal to the manifests. The applier's credential verification is now scoped to the Deployments actually planned, so a cost-monitoring-off deployment no longer warns about cost-monitor.
- `tests/test_network_policies_manifest.py` rewritten for the new posture; applier tests for the prune inventory and the legacy sweep; config tests for every new `cdk.json` key and its range; a cdk.json test-matrix entry (`vpc-interface-endpoints-enforcement-off`) that synthesizes interface endpoints with enforcement off; unit tests for the shared `LeaseElector` and the dispatcher's leadership gating; graceful-shutdown budget tests for every FastAPI entrypoint.
- Kind e2e (`integration:kind:cluster-e2e`): manifests are rendered by a small Python renderer with startable health-monitor/manifest-processor configuration, the manifest-processor HPA gate is rendered ON and read back, `rollout status` is required for the health-monitor and manifest-processor too (they used to crash-loop on stub values and were scaled to 0 before anyone looked), every PDB's `maxUnavailable` is read back, and the NetworkPolicy probes now prove verdicts both ways: cross-namespace into `gco-system` blocked, same-namespace in `gco-jobs` reachable, cross-namespace into `gco-jobs` blocked, the inference-monitor metrics port reachable from another namespace and its other port blocked from the same pod.
- Live release validation gains two actions, run after every workload action so their evidence describes services that have carried real traffic: `platform-workloads` (every `gco-system` Deployment converged at its current generation with zero container restarts, each PDB allowing a disruption, the configured HPAs present and able to scale, and the `amazon-vpc-cni` switch carrying the configured value) and `network-posture` (run-labelled, digest-pinned probe Jobs dialing real listeners, the live inference-monitor and an AWS HTTPS endpoint from `gco-jobs` and an unpoliced namespace; reachability and deny verdicts must both match, and every Job is deleted before the action returns). Both use the inference action's tunnelled kubectl session; preflight now requires the Session Manager plugin for any of the three cluster-facing actions.

## What was deliberately NOT changed

- Readiness stays shallow (process up, background task alive). Gating `/readyz` on DynamoDB or Secrets Manager would turn a dependency blip into a fleet-wide unready event behind the ALB.
- Application containers keep binding loopback and probing through a `python` exec: the plaintext port is never network-reachable by construction, and that property is worth more than kubelet-native probes. The exec itself got cheap after the sixth live run (`-I -S`, bare socket, no `urllib`), and liveness now restarts only after a minute of continuous failure; see the follow-up section.
- No digest pinning: CDK exposes the asset tag, not the digest, and the tag is a content hash.
- `gco-jobs` egress is not restricted to VPC endpoints only: no interface endpoints exist by default, the global Region's tables and bucket are cross-Region, and model/package downloads need the internet. 443-anywhere plus DNS plus in-VPC is the honest floor; the deny on everything else is proven live.
- The inference-monitor keeps the shared workload IAM role; splitting a least-privilege role needs an API inventory and a live validation run — listed below.
- cost-monitor stays a plain-HTTP, NetworkPolicy-scoped internal Service; only the manifest-processor may reach it.

## Observations, not fixed here

- inference-monitor should have its own IAM role instead of `gco-service-account`'s; requires inventorying its DynamoDB/SSM calls and a live run.
- The manifest-processor API handlers call the Kubernetes client synchronously inside async routes; moving them to `asyncio.to_thread` would let one replica serve concurrent requests and would make CPU a more meaningful autoscaling signal.
- The auth middleware's replay-nonce set is per process; replicas do not share it.
- Per-replica CloudWatch publication from the health-monitor doubles datapoints; harmless for the statistics consumed, noted for anyone reading raw sample counts.
- The live-validation runner re-runs the final inventory as its guaranteed post-run re-check and records that pass with a zero duration over the action's own result (the run's `final-inventory` took 9.7 min and the report table shows 0 s). Cosmetic; left for a follow-up so the validated SHA stays the one that ran.
- On one validation run the node hosting the health-monitor leader spent most of both vCPUs in kernel time for twenty minutes while every other node sat under 20%; the node's kubelet and containerd journals name no cause. The platform now rides it out (see the follow-up section), but the source of that pressure on an Auto Mode node is not understood.

## Risks

- `imagePullPolicy: IfNotPresent` relies on the asset tag being immutable; CDK's content-hash tags are, and the kind job already rendered it this way.
- Switching the policy controller on is the one change here that alters live traffic: everything the platform services and the shipped smoke jobs need is admitted by the corrected rules and proven by the kind probes and the new live `network-posture` action, but a workload that relied on the policies being inert (anything reaching a `gco-jobs` pod from another namespace, or leaving `gco-jobs` on a non-443 port to a non-VPC address) will now be blocked. `eks_cluster.network_policy_enforcement: false` restores the old behaviour.
- Renamed nothing, so no pod restarts beyond the normal deployment-timestamp roll; the inference-monitor rolls once more than before on config-only deploys.
- The optional HPA is off by default; the prune inventory removes it when turned off again.
- **Updating an existing deployment replaces four S3 buckets** (cost-report, cluster-shared, regional-shared, analytics-studio): CloudFormation treats the move from explicit names to generated ones as a replacement, so their contents must be copied out first ([ADR-0005](docs/adr/0005-cloudformation-generated-s3-bucket-names.md)).
- A stuck platform container is now restarted after a minute of failed liveness probes rather than 45 seconds; the wider window is what keeps a CPU-starved node from restarting healthy containers.

## Testing

- [x] Targeted `pytest` locally: every test file touched by the branch plus the service, manifest, applier, config-loader, live-validation and docs-gate suites (~4,000 tests), the regional stack suite in class batches, kubeconform online (`GCO_KUBECONFORM_VALIDATION=1`), and 100% line+branch coverage on every new module. The CDK synthesis matrix and the full sharded suite run in CI only (they are too large for the local disk).
- [x] `cdk synth` is exercised by the regional stack tests (the matrix entry with interface endpoints and enforcement off synthesizes in CI).
- [x] New tests added for new behavior (`tests/README.md` rows for every new file).
- [x] Ran the change against a real AWS account — live release validation below.
- [x] `ruff format --check`, `ruff check`, strict `mypy` (both CI invocations), `actionlint`, `yamllint`, `markdownlint`, and `checkov` against the new probe manifests all clean.

## Live release validation

- [ ] Not required
- [x] Required and completed locally with `--actions all` for this exact SHA; a sanitized `PASSED` summary comment (run ID, SHA, per-action statuses) was posted on the pull request
- [ ] Required but pending explicit validation-account and KMS-deletion authorization

**Applicability rationale:** this PR changes deployed infrastructure (VPC endpoints, Pod Identity associations, dropped ECR repositories), every platform Deployment's pod spec, and — for the first time — actually enforces the NetworkPolicies on the live cluster. A full `--actions all` run, including the two new `platform-workloads` and `network-posture` actions, completed with status `PASSED` on the final SHA (`c7af22902fc4`, run `pr375-release-c7af2290-20260913T031429Z`); six earlier full runs on this branch each stopped on a finding that was fixed rather than retried (see the follow-up section).

**Sanitized validation summary or comment link:** https://github.com/aws-solutions-library-samples/global-capacity-orchestrator-on-aws/pull/375#issuecomment-5651732284

## Checklist

- [x] Documentation updated (`docs/ARCHITECTURE.md`, `docs/CUSTOMIZATION.md`, `docs/LIVE_RELEASE_VALIDATION.md`, `docs/CLI.md`, `docs/adr/0005-cloudformation-generated-s3-bucket-names.md`, `docs/COST_MONITORING.md`, `docs/KUBERAY.md`, scheduler docs, `lambda/kubectl-applier-simple/manifests/README.md`, `gco/services/README.md`, `scripts/live_release_validation/README.md`, `.github/CI.md`, `cdk.json` comments)
- [x] `requirements-lock.txt` unchanged (`pyproject.toml` untouched)
- [x] Changes align with the architecture described in `docs/ARCHITECTURE.md` (its Network Security section was updated to match)


Jmevorach:
Live release validation ran locally with `--actions all` against the dedicated, disposable validation account for this exact SHA. The full reports stay on the operator's machine; this is the sanitized summary.

- Run ID: `pr375-release-c7af2290-20260913T031429Z`
- Commit: `c7af22902fc43aa3dd63f4505fbf5084b99a1b01`
- Overall report status: **PASSED** (16 actions, statuses: passed)
- Cleanup completed: True; every run-owned stack absent afterwards: True; protected-baseline differences: 0
- Final inventory accepted, with evidence, 7 KMS keys already in their scheduled-deletion window; 2 ECR repositories intentionally retained; 4 Tagging-API records of VPC endpoints EC2 had already deleted; 1 EFS automatic-backup recovery points.

|---|---|---|
| `preflight` | passed | 39 s |
| `baseline` | passed | 8.8 min |
| `deploy` | passed | 42.4 min |
| `topology` | passed | 9.5 min |
| `inference` | passed | 32.9 min |
| `policy` | passed | 2 s |
| `api` | passed | 24 s |
| `sqs` | passed | 44 s |
| `central-queue` | passed | 36 s |
| `schedulers` | passed | 1.8 min |
| `opencost` | passed | 3 s |
| `convergence` | passed | 21 s |
| `platform-workloads` | passed | 2.5 min |
| `network-posture` | passed | 4.4 min |
| `destroy` | passed | 79.4 min |
| `final-inventory` | passed | 9.7 min¹ |

¹ The action itself took 9.7 min; the runner then re-ran the inventory as its guaranteed post-run re-check and the report records that pass's timestamp, which is why the report table shows 0 s for it (noted under "Observations, not fixed here").

PR #382, opened 2026-09-16
feat: gco upgrade, zero-Region topologies, and a planning-reference guard

Jmevorach:
## Summary

Three related changes, one per commit:

**1. `chore:` remove leaked planning-document references; widen the guard.** The MkDocs hook, `mkdocs.yml`, the wiki guard, one production docstring, one CDK comment and a handful of tests still pointed at internal planning artifacts (a spec directory, "Requirement 4.2", "design §2.3", "task 16.9", "Validates: Property N"). Every pointer is rewritten in plain terms. `tests/test_no_spec_references.py` could not have caught them — it still listed the pre-rename `mcp/` package (so `gco_mcp/` was never scanned), skipped `wiki/`, `.github/` and every root file, and matched only fixed substrings. It now walks every shipped tree plus the repository root, adds labelled regexes for the numbered breadcrumb shapes, pins that every scanned directory exists, and self-tests each pattern against the known leaks and against legitimate domain vocabulary (Kubernetes `spec`, "Add to Kiro", AWS "requirements doc").

**2. `feat:` allow zero workload Regions and a control-plane-preserving teardown.** `deployment_regions.regional` may now be `[]`: a control-plane-only topology where only the global, API Gateway and monitoring stacks synthesize. Every per-Region surface degrades cleanly instead of failing — the API Gateway stack emits its aggregator grants only when there is a Region to describe (an identity statement without resources fails synth); the backend-TLS manager accepts an empty `Regions` list, keeps the root CA and retires the previous leaf certificates; the cross-region aggregator answers with empty results rather than a 503; the monitoring dashboard replaces its per-Region sections with a note (a single-value widget with no metrics is rejected by CloudWatch at deploy). A run-scoped context key, `gco:control-plane-only=true`, synthesizes the same shape without editing `cdk.json`. `destroy_orchestrated(keep_control_plane=True)` — exposed as `gco stacks destroy-all --keep-control-plane` and on the `destroy_all` MCP tool — uses it to update the monitoring stack in place so it stops referencing the regional stacks, then removes only the bridges and regional stacks, skipping the full-teardown sweeps (image-registry preflight, backup-vault purge, bastion IAM retirement, traffic-dial purge). `gco stacks regions remove` may remove the last Region and says what that means; `gco stacks access` refuses to guess a Region when none is configured. The synthesis matrix gains `zero-regions`, `control-plane-only-context` and `feature-enabled-overrides-context` entries; the knob-coverage guard now resolves constant-named `try_get_context` reads, which is what surfaced the last one.

**3. `feat:` `gco upgrade`.** Runs inside a git checkout and, in one pass: fetches the release tags and checks out the highest `vMAJOR.MINOR.PATCH` (or `--ref`) with `cdk.json` preserved byte-for-byte (modified or fork-committed); refreshes the local install (`pip install -e .` for an editable CLI with a `uv pip` fallback, `npm ci` for a checkout-local CDK toolchain, a rebuild of the `gco-dev` image when it exists); scales the workload tier to zero with the control plane standing; then runs `deploy_orchestrated` so the global and API Gateway stacks update in place and the regional stacks, their bridges and the monitoring stack are recreated on the new release. It prints the plan, names every stack it will destroy, spells out the data that goes with them, and asks for the project name to be typed unless `-y` is given. `--check` reports without changing anything; `--skip-checkout` / `--force` rerun only the stack cycle; failures name the phase and the resume command; machine modes receive one document. `docs/UPGRADING.md` is the procedure (back regional data up to the cluster-shared bucket, run, restore); CLI.md, the module READMEs, the docs index, the MCP docs catalogue and the wiki point at it. There is deliberately no MCP tool: the command replaces the checkout the MCP server runs from.

A fourth commit regenerates the code flowcharts for the charted sources these changes touched.

## Type of change

- [x] `chore:` Maintenance (dep bumps, etc.)

## Testing

- [x] `pytest tests/` passes locally — full core suite (16,980 tests, `-n 8`); every changed or added module is at 100% line+branch coverage. The only local misses were environmental (`/dev/fd` traversal tests skipped on macOS, `test_kubectl_helpers_extended` fails identically on `main` locally, `scripts/accelerator_catalog.py` is covered by its dedicated CI job).
- [x] `cdk synth` succeeds (if CDK code changed) — `tests/test_cdk_synthesis_matrix.py` including the three new matrix entries; `tests/test_nag_compliance.py` subset unchanged.
- [x] New tests added for new behavior — `tests/test_upgrade.py` (59, real temporary git repositories for the checkout semantics), `tests/test_cli_upgrade.py` (34), keep-control-plane teardown and monitoring-detach tests in `tests/test_stacks_destroy_hardening.py`, zero-region tests across config loader, managed config, api-gateway stack, TLS manager and aggregator handlers, plus the guard's own pattern self-test.
- [x] `ruff format --check`, `ruff check`, `mypy` (strict set), `bandit -r . --severity-level medium`, `yamllint`, `markdownlint-cli2` (128 files), `mkdocs build --strict` all clean.
- [ ] Ran the change against a real AWS account (describe below) — in progress in the validation account; results will be posted as a comment.

## Live release validation

- [x] Not required (explain the applicability decision below)

**Applicability rationale:** The deployed templates change only for the new zero-Region topology (guards that remove resources when there is nothing to reference); the default topology synthesizes identically (asserted by the synthesis matrix). The new command composes the existing, already-validated `destroy_orchestrated` / `deploy_orchestrated` paths. A live scale-to-zero → scale-up → `gco upgrade --skip-checkout` cycle is being run in the validation account regardless and its sanitized outcome will be posted on this PR.

**Sanitized validation summary or comment link (never post the full reports or `checkpoint.json`; they carry account IDs, ARNs, and endpoint URLs):** to follow as a PR comment.

## Checklist

- [x] Documentation updated (README, `docs/`, inline docstrings) as needed
- [x] `requirements-lock.txt` regenerated if `pyproject.toml` changed (unchanged)


Jmevorach:
## Live validation summary

Everything below was exercised end to end in a scratch AWS account (control plane in one Region, one workload Region), driven by this branch's CLI against real CloudFormation, EKS Auto Mode, API Gateway and ECR. Account ids, ARNs and endpoint URLs are omitted.

### 1. Zero workload Regions (control plane only)

With `deployment_regions.regional: []`:

- `gco stacks deploy-all` created **3/3** stacks — global, API Gateway, monitoring — no regional stack, no regional API bridge.
- The monitoring dashboard synthesized and deployed with the "no workload Regions" note widget instead of per-Region sections (an empty `SingleValueWidget` is rejected by CloudWatch, which is why the note exists).
- The API Gateway aggregator deployed without the per-Region IAM statements.
- `gco jobs list --all-regions` returned a successful, empty answer: `Total jobs: 0 / Regions queried: 0 / Regions successful: 0` — the aggregator answers rather than failing when there is nothing to fan out to.

### 2. Scaling back up

`gco stacks regions add <region>` followed by `deploy-all` brought the topology to **5/5**: the regional stack and its API bridge were created, and the global / API Gateway / monitoring stacks were updated in place (monitoring re-attached its per-Region alarms and cross-Region reader).

### 3. `gco upgrade`

`gco upgrade --check` printed the plan without touching anything: current version → target tag, the local-install steps, the destroy order (bridge before regional stack), the in-place control-plane list, and the data-loss warning pointing at `docs/UPGRADING.md`. The machine-readable form (`--output json`) reported the same plan, with `cdk.json` listed under the preserved-but-modified files.

The real run (`--skip-checkout`, so the shared editable install was left alone) completed with exit 0 and, in order:

- **Phase 1** — monitoring updated in place so it no longer references the regional stacks (the cost-report bucket and its history survive), then the regional API bridge and the regional stack destroyed. The teardown's own cleanup reported the empty EKS security group, two orphaned EBS volumes left by in-cluster Prometheus/Alertmanager PVCs, and 18 implicit log groups.
- **Phase 2** — global and API Gateway updated in place, the regional stack and bridge recreated (third-party image mirroring correctly skipped the already-mirrored tags), monitoring re-attached.

Final state: `Upgrade to <tag> complete`, `destroyed and recreated: <bridge>, <regional>`, `deployed: global, api-gateway, regional, bridge, monitoring`.

Post-upgrade verification: `gco status` showed all five stacks healthy, the queue reachable with zero depth, and the recreated ALB's three target groups all healthy. `gco jobs list --all-regions` initially returned a gateway timeout while the freshly created cluster's API pods were still starting, then answered normally (`Regions queried: 1 / Regions successful: 1`) a couple of minutes later — cold start, not a regression.

### 4. Teardown

`gco stacks destroy-all` removed every stack in both Regions; the deployment's `cdk.json` was restored to its committed contents afterwards.

### Not covered live

Multi-Region (>1 workload Region) upgrade fan-out, `--parallel`, and the full checkout path (`git checkout` of a newer tag plus `pip install -e .` / `npm ci` / dev-image rebuild) — the branch is ahead of every existing tag, so the checkout step had nothing newer to move to and was skipped deliberately. Those paths are covered by the unit suite.

